PR #2, opened 2020-09-25
Draft: FilePath.ComponentView

milseman:
Draft: Add FilePath.ComponentView, a Bidi RRC of components
    
Adds in testing infrastructure and collectiony tests, as well
as ad-hoc testing.

### DIFF
--- a/Sources/System/FilePathComponents.swift
+++ b/Sources/System/FilePathComponents.swift
@@ -1,0 +1,345 @@
+/*
+ This source file is part of the Swift System open source project
+
+ Copyright (c) 2020 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+// @available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
+extension FilePath {
+  public struct Component: Hashable {
+    // NOTE: For now, we store a slice of FilePath's storage representation. We'd like to
+    // have a small-slice representation in the future since the majority of path
+    // components would easily fit in the 3 words of storage.
+    //
+    internal var slice: FilePath.Storage.SubSequence
+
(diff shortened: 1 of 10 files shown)


Sources/System/FilePathComponents.swift
```diff
@@ -0,0 +1,345 @@
...
+
+    // Explicitly add a trailing separator if
+    //   not at end and next character is not a separator
+    let atEnd = upperBound == endIndex
+    let trailingSeparator = !atEnd && !isSeparator(path.bytes[upperBound._storage])
+
+    // Explicitly add a preceding separator if
+    //   replacing front with absolute components (unless redundant by trailing separator),
+    //   preceding character is not a separator (which implies at the end)
+    let atStart = lowerBound == startIndex
```
milseman, line 279: TODO: Account for `//` special case...

> Multiple successive <slash> characters are considered to be the same as one <slash>, except for the case of exactly two leading <slash> characters.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_271

milseman: For POSIX, it seems to denote implementation defined path resolution:

> A pathname consisting of a single <slash> shall resolve to the root directory of the process. A null pathname shall not be successfully resolved. If a pathname begins with two successive <slash> characters, the first component following the leading <slash> characters may be interpreted in an implementation-defined manner, although more than two leading <slash> characters shall be treated as a single <slash> character.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13

lorentey: Luckily, all signs indicate that `//` is special cased on neither Darwin nor Linux, so there is no need for us to do anything about it. It's okay to treat `//` as a synonym of `/` for the purposes of canonicalization and component extraction.

milseman:
This work is superseded by FilePath syntactic operations: https://github.com/apple/swift-system/pull/14

PR #6, opened 2020-10-07
Initial port to Windows

compnerd:
This adds enough shims and tweaks the types sufficiently to allow
building System on Windows.

### DIFF
--- a/Sources/System/Errno.swift
+++ b/Sources/System/Errno.swift
@@ -229,6 +229,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @available(*, unavailable, renamed: "badAddress")
   public static var EFAULT: Errno { badAddress }
 
+#if !os(Windows)
   /// Not a block device.
   ///
   /// You attempted a block device operation on a nonblock device or file.
@@ -240,6 +241,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   @available(*, unavailable, renamed: "notBlockDevice")
   public static var ENOTBLK: Errno { notBlockDevice }
+#endif
 
   /// Resource busy.
   ///
@@ -360,6 +362,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @available(*, unavailable, renamed: "tooManyOpenFiles")
   public static var EMFILE: Errno { tooManyOpenFiles }
 
+#if !os(Windows)
   /// Inappropriate control function.
   ///
   /// You attempted a control function
@@ -388,6 +391,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   @available(*, unavailable, renamed: "textFileBusy")
   public static var ETXTBSY: Errno { textFileBusy }
+#endif
 
   /// The file is too large.
   ///
@@ -630,6 +634,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @available(*, unavailable, renamed: "socketTypeNotSupported")
   public static var ESOCKTNOSUPPORT: Errno { socketTypeNotSupported }
 
+#if !os(Windows)
   /// Not supported.
   ///
   /// The attempted operation isn't supported
@@ -642,6 +647,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   @available(*, unavailable, renamed: "notSupported")
   public static var ENOTSUP: Errno { notSupported }
+#endif
 
   /// Protocol family not supported.
   ///
@@ -1123,6 +1129,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   public static var EDEVERR: Errno { deviceError }
 #endif
 
+#if !os(Windows)
   /// Value too large to be stored in data type.
   ///
   /// A numerical result of the function
@@ -1135,6 +1142,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   @available(*, unavailable, renamed: "overflow")
   public static var EOVERFLOW: Errno { overflow }
+#endif
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
   /// Bad executable or shared library.
@@ -1199,6 +1207,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @available(*, unavailable, renamed: "canceled")
   public static var ECANCELED: Errno { canceled }
 
+#if !os(Windows)
   /// Identifier removed.
   ///
   /// An IPC identifier was removed while the current process was waiting on it.
@@ -1223,6 +1232,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   @available(*, unavailable, renamed: "noMessage")
   public static var ENOMSG: Errno { noMessage }
+#endif
 
   /// Illegal byte sequence.
   ///
@@ -1252,6 +1262,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   public static var ENOATTR: Errno { attributeNotFound }
 #endif
 
+#if !os(Windows)
   /// Bad message.
   ///
   /// The message to be received is inappropriate
@@ -1350,6 +1361,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   @available(*, unavailable, renamed: "timeout")
   public static var ETIME: Errno { timeout }
+#endif
 
   /// Operation not supported on socket.
   ///
@@ -1411,6 +1423,7 @@ extension Errno {
   public static var ENOPOLICY: Errno { noSuchPolicy }
 #endif
 
+#if !os(Windows)
   /// State not recoverable.
   ///
   /// The corresponding C error is `ENOTRECOVERABLE`.
@@ -1430,6 +1443,7 @@ extension Errno {
   @_alwaysEmitIntoClient
   @available(*, unavailable, renamed: "previousOwnerDied")
   public static var EOWNERDEAD: Errno { previousOwnerDied }
+#endif
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
   /// Interface output queue is full.
--- a/Sources/System/WindowsPlatformConstants.swift
+++ b/Sources/System/WindowsPlatformConstants.swift
@@ -1,0 +1,328 @@
+/*
+ This source file is part of the Swift System open source project
+
+ Copyright (c) 2020 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+#if os(Windows)
+
+@_alwaysEmitIntoClient
+internal var _EPERM: CInt { 1 }
+
+@_alwaysEmitIntoClient
+internal var _ENOENT: CInt { 2 }
+
+@_alwaysEmitIntoClient
+internal var _ESRCH: CInt { 3 }
+
+@_alwaysEmitIntoClient
+internal var _EINTR: CInt { 4 }
+
+@_alwaysEmitIntoClient
+internal var _EIO: CInt { 5 }
+
+@_alwaysEmitIntoClient
+internal var _ENXIO: CInt { 6 }
+
+@_alwaysEmitIntoClient
+internal var _E2BIG: CInt { 7 }
+
+@_alwaysEmitIntoClient
+internal var _ENOEXEC: CInt { 8 }
+
+@_alwaysEmitIntoClient
+internal var _EBADF: CInt { 9 }
+
+@_alwaysEmitIntoClient
+internal var _ECHILD: CInt { 10 }
+
+@_alwaysEmitIntoClient
+internal var _EAGAIN: CInt { 11 }
+
+@_alwaysEmitIntoClient
+internal var _ENOMEM: CInt { 12 }
+
+@_alwaysEmitIntoClient
+internal var _EACCES: CInt { 13 }
+
+@_alwaysEmitIntoClient
+internal var _EFAULT: CInt { 14 }
+
+@_alwaysEmitIntoClient
+internal var _EBUSY: CInt { 16 }
+
+@_alwaysEmitIntoClient
+internal var _EEXIST: CInt { 17 }
+
+@_alwaysEmitIntoClient
+internal var _EXDEV: CInt { 18 }
+
+@_alwaysEmitIntoClient
+internal var _ENODEV: CInt { 19 }
+
+@_alwaysEmitIntoClient
+internal var _ENOTDIR: CInt { 20 }
+
+@_alwaysEmitIntoClient
+internal var _EISDIR: CInt { 21 }
+
+@_alwaysEmitIntoClient
+internal var _EINVAL: CInt { 22 }
+
+@_alwaysEmitIntoClient
+internal var _ENFILE: CInt { 23 }
+
+@_alwaysEmitIntoClient
+internal var _EMFILE: CInt { 24 }
+
+@_alwaysEmitIntoClient
+internal var _ENOTIFY: CInt { 25 }
+
+@_alwaysEmitIntoClient
+internal var _EFBIG: CInt { 27 }
+
+@_alwaysEmitIntoClient
+internal var _ENOSPC: CInt { 28 }
+
+@_alwaysEmitIntoClient
+internal var _ESPIPE: CInt { 29 }
+
+@_alwaysEmitIntoClient
+internal var _EROFS: CInt { 30 }
+
+@_alwaysEmitIntoClient
+internal var _EMLINK: CInt { 31 }
+
+@_alwaysEmitIntoClient
+internal var _EPIPE: CInt { 32 }
+
+@_alwaysEmitIntoClient
+internal var _EDOM: CInt { 33 }
+
+@_alwaysEmitIntoClient
+internal var _ERANGE: CInt { 34 }
+
+@_alwaysEmitIntoClient
+internal var _EDEADLK: CInt { 36 }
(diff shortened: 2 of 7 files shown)


compnerd:
CC: @kylemacomber @milseman

Sources/System/WindowsPlatformConstants.swift
```diff
@@ -0,0 +1,330 @@
...
+
+ Copyright (c) 2020 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+#if os(Windows)
+
+import ucrt
```
milseman, line 12: What is this import used for in this file? We also want to make sure that System doesn't pull in any other modules, so make sure all imports from the System module are `@_implementationOnly`

Sources/SystemInternals/Exports.swift
```diff
@@ -28,6 +30,18 @@ public var system_errno: CInt {
   get { Darwin.errno }
   set { Darwin.errno = newValue }
 }
+#elseif os(Windows)
+public var system_errno: CInt {
+  get {
+    var value: CInt = 0
+    // TODO(compnerd) handle the error?
```
milseman, line 37: What is there to handle?

Sources/SystemInternals/Exports.swift
```diff
@@ -36,47 +50,129 @@ public var system_errno: CInt {
 #endif
 
 public func system_open(_ path: UnsafePointer<CChar>, _ oflag: Int32) -> CInt {
+#if os(Windows)
+  var fh: CInt = -1
+  _ = _sopen_s(&fh, path, oflag, _SH_DENYNO, _S_IREAD | _S_IWRITE)
```
milseman, line 55: Is this the rough equivalent of POSIX open? E.g., what is the difference between a Windows file handle and a POSIX file descriptor?

Sources/SystemInternals/Exports.swift
```diff
@@ -36,47 +50,129 @@ public var system_errno: CInt {
...
+  return fh
+#else
   open(path, oflag)
+#endif
 }
 
+#if os(Windows)
+public func system_open(
+  _ path: UnsafePointer<CChar>, _ oflag: Int32, _ mode: CInt
+) -> CInt {
```
milseman, line 65: I wonder if we should try to unify the interface around trivial type casts

Sources/SystemInternals/Exports.swift
```diff
@@ -36,47 +50,129 @@ public var system_errno: CInt {
...
+  ovlOverlapped.OffsetHigh = DWORD(UInt32(offset >> 32) & 0xffffffff)
+  ovlOverlapped.Offset = DWORD(UInt32(offset >> 0) & 0xffffffff)
+
+  var nNumberOfBytesRead: DWORD = 0
+  if !ReadFile(hFile, buf, DWORD(nbyte), &nNumberOfBytesRead, &ovlOverlapped) {
+    let _ = GetLastError()
+    // TODO(compnerd) map windows error to errno
+    return Int(-1)
+  }
+  return Int(nNumberOfBytesRead)
```
milseman, line 120: Could you explain more what is going on here? Does Windows not support pread/pwrite? We will have to decide on a case-by-case basis what should be handled by availability and what should be handled by implementing the functionality ourselves (but this doesn't look too bad). This is definitely worthy of being extracted into a helper function and given a good name (and comments!) since it's non-obvious to a non-Windows contributor what's going on.

milseman (commented):
Just some initial comments and questions. Glad to see you working on this!

Sources/System/WindowsPlatformConstants.swift
compnerd: Leftovers :-( I realized that later that the values are being replicated rather than imported.

Sources/SystemInternals/Exports.swift
compnerd: `_get_errno` returns an `errno_t`.  It can fail in case of a PEB/TEB corruption or the parameter being invalid.

Sources/SystemInternals/Exports.swift
compnerd: Yes, this is roughly the equivalent of the POSIX `open`.

I don't think that its valuable to try to enumerate every single difference.  What type of differences are you really looking for?

But for a taste, there are a large number of differences between a file handle and a POSIX file.  The POSIX file is a user mode only, process specific construct.  It _cannot_ be shared across processes.  It is a mapping from the kernel side representation (`HANDLE`) and an application specific view.  It is maintained by means of a lookup table in the application's address space by the C library.  Additionally, it has a comparatively restricted set of operations available to it.

Sources/SystemInternals/Exports.swift
compnerd: I was considering that myself, but wasn't sure if you would be okay with choosing the minimal type across the interfaces.

Sources/SystemInternals/Exports.swift
compnerd: Yes, there is no `pread` or `pwrite` function that is available in the C library.  However, the operation itself _is_ supported, but you need to use the Win32 APIs to do that.  Basically, I have to convert the file descriptor to the kernel/win32 representation (`HANDLE`) and then I can perform the operation via the Win32 API (`ReadFile` or `WriteFile`) as I don't see the value in going directly to the syscall interface (`NtReadFile` or `NtWriteFile`) as they are more complicated to use (it would directly map to the kernel call `ZwReadFile` or `ZwWriteFile`).

jakepetroules:
@compnerd How do you plan to handle file paths in SwiftSystem? You probably know what I'm referring to, but to spell it out for the record: 8-bit file paths on Windows are not UTF-8 like on POSIX (except on Windows 10 and even then only in certain cases). Should SwiftSystem go through the Win32 APIs instead, in order to ensure proper Unicode support?

compnerd:
@jakepetroules thats something that I did bring up with @milseman and the plan is to have a representation which is platform dependent AIUI.  So that would enable the paths to handle the proper conversion at the right time.

milseman:
Right, so on nix you'd have an initializer taking a collection of bytes in some platform encoding (e.g. UTF-8) and on Windows you'd have an init taking (presumably) a collection of UInt16s. Both platforms can form paths from strings and both platforms can (with potential encoding error correction) go to a string. The storage representation of `FilePath` is hidden and opaque (even in the ABI).

Sources/System/Errno.swift
```diff
@@ -1187,6 +1195,19 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
...
+  /// The scheduled operation was canceled.
+  ///
+  /// The corresponding C error is `ECANCELED`.
+  @_alwaysEmitIntoClient
+  public static var canceled: Errno { Errno(_ECANCELLED) }
+
+  @_alwaysEmitIntoClient
+  @available(*, unavailable, renamed: "canceled")
+  public static var ECANCELLED: Errno { canceled }
+#else
```
benrimmington, line 1210: [`ECANCELED`][] appears to be supported, possibly by defining [`_CRT_NONSTDC_NO_WARNINGS`][].

Otherwise, does the above comment need to change?

```diff
-  /// The corresponding C error is `ECANCELED`.
+  /// The corresponding C error is `ECANCELLED`.
```

[`ECANCELED`]:                <https://docs.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=vs-2019>
[`_CRT_NONSTDC_NO_WARNINGS`]: <https://docs.microsoft.com/en-us/cpp/c-runtime-library/compatibility?view=vs-2019>

Sources/SystemInternals/Exports.swift
compnerd: If we want to do that, I think that doing a follow up might be the best way forward.

Sources/System/Errno.swift
compnerd: Thanks for catching that!  That should have been `canceled`.  It doesn't help that I have a tendency to spell it as "cancelled".  I think it is more valuable to have the common spelling here rather than preserve the difference between the spelling that Windows adopts and Unix adopts.

Sources/SystemInternals/Exports.swift
```diff
@@ -36,47 +50,129 @@ public var system_errno: CInt {
...
 public func system_lseek(
   _ fd: Int32, _ off: off_t, _ whence: Int32
 ) -> off_t {
-  lseek(fd, off, whence)
+#if os(Windows)
+  // TODO(compnerd) should this use _lseeki64 to get larget file offsets?
+  return _lseek(fd, off, whence)
+#else
+  return lseek(fd, off, whence)
+#endif
```
milseman, line 134: To reduce the number of `#if/#else` inside of these functions, should Windows have a `internal let lseek = _lseek`, etc?

milseman:
This LGTM after I merge mocking at https://github.com/apple/swift-system/pull/8. I apologize in advance that it will cause you some merge conflicts around the `system_*` functions. Hopefully we can get most of the `#if/#else` out of there.

milseman:
@compnerd sorry for the conflict, it's worth moving these sys call wrappers into their own file. The good news is that I was able to structure the mocking-related code such that there is less boilerplate and it all appears at the very top of the function, so your code shouldn't need much modification.

compnerd:
That change introduced more than just conflicts :/  The threading model doesn't work for Windows - there is no `pthread_t` on Windows.  I think that I can rebase the change, and we should merge it in the broken state, I'll redo the threading in a follow up.

milseman:
For Windows, would you just use `TlsGet/SetValue`? This isn't System API, it's in the internals, so it's just an implementation detail. We will definitely want to expose System API for TLS in the future, but that's further down the road and would need a more comprehensive design.

milseman:
If you can show me the Windows calls you'd like use and an example of how to use them, I can refactor the mocking code to accommodate Windows better.

Sources/SystemInternals/Exports.swift
compnerd: There is a `lseek` as well, so I don't think that we should be shadowing that.

compnerd:
@milseman, @kylemacomber - this part is rebased, I'll do a follow up for the threading

compnerd:
@milseman - lets discuss the threading in #9.

Sources/SystemInternals/Syscalls.swift
```diff
@@ -35,11 +38,19 @@ private func mockImpl(
...
+    _ = ucrt._set_errno(e)
+#else
     errno = e
+#endif
     return -1
   case .counted(let e, let count):
     assert(count >= 1)
+#if os(Windows)
+    _ = ucrt._set_errno(e)
+#else
```
milseman, line 51: I'm definitely in favor of making a helper function to set errno if it means we can avoid scattering `#if/#else` throughout function bodies. But that can be a subsequent PR.

milseman (approved):
This LGTM, but I would like to do some cleanup. If there is something blocked by this, that can happen on a subsequent PR. If not, then it might be cleaner to do it here.

1. Can Windows just define `open`, `pwrite`, `lseek`, etc? That would keep the `system_foo` bodies clean and straight forward. It's an unfortunate amount of boilerplate we have to have in System while we wait on compiler features (or we do codegen ourselves), so ideally the bodies would be simple and uniform.

2. We'd like to unify the types for `system_write` and `system_read` interfaces as much as possible. I'm worried It might be too easy to break Windows when modifying the System module in what seems to be trivial changes. We can unify on the wider type.

compnerd:
1. I tried to do that, but the you cannot do a typealias locally.  I don't want to make it shadow the existing functions with the same name.  Do you have another suggestion on how to do that?  I suppose I can create `sys...` variants of the functions and use that.

2. Why would we unify on the wider type?  Doesn't that mean that some inputs become invalid?  Shouldn't it be the narrower type?

milseman:
One approach might be to add a `WindowsSyscalls.swift` that has something like:

```swift
internal let lseek = _lseek
internal func open(...) {
  var handle: ...
  _open(&handle, ...)
  ...
  return handle
}
```

The advantage of this is it eliminates the `#if` in the bodies of the forwarding methods, which eventually will likely number in the hundreds (and we may end up using code generation for). This disadvantage is needing one more jump-to-definition hop when inspecting the source code. This also might be a good place to put the type widening/narrowing. WDYT?

milseman:
For "widening", what I mean is effectively moving the extending/truncating code from `FileDescriptor.read` into SystemInternals, so that instead of 

```swift
// FileOperations.swift
extension FileDescriptor {
  func read(...) {
    Int(system_read(self.rawValue, buffer.baseAddress, numericCast(buffer.count)))
  }
}
```

We'd have:

```swift
// WindowsSyscalls.swift
#if os(Windows)
func read(_ fd: Int32, _ buf: UnsafeMutableRawPointer!, _ nbyte: Int) -> Int {
  Int(_read(fd, buf, numericCast(nbyte)))
}
#endif
```

being semantically equivalent and removing the cognitive burden from anyone touching other parts of System.

milseman:
I could take a stab at this if you like. I wouldn't be able to test it (or even compile it), but it might be a good proof of concept.

Sources/SystemInternals/Syscalls.swift
```diff
@@ -74,9 +85,27 @@ public func system_open(_ path: UnsafePointer<CChar>, _ oflag: Int32) -> CInt {
...
+public func system_open(
+  _ path: UnsafePointer<CChar>, _ oflag: Int32, _ mode: CInt
+) -> CInt {
+#if ENABLE_MOCKING
+  if mockingEnabled { return mock(String(cString: path), oflag, mode) }
+#endif
+  var fh: CInt = -1
+  _ = _sopen_s(&fh, path, oflag, _SH_DENYNO, _S_IREAD | _S_IWRITE)
+  return fh
+}
```
milseman, line 107: `mode` is not used, so why do we care to type it as a `CInt`? For Windows, would we not have a permissions parameter for `FileDescriptor.open`?.

milseman:
https://github.com/apple/swift-system/pull/11 is a draft that shows what I mean, let me know what you think.

Sources/SystemInternals/Syscalls.swift
compnerd: There is no `mode_t` type, and I don't think that you can use an internal typealias on a public func.

compnerd:
Hmm, as long as we can ensure that the function gets flattened, I think that might not be a bad idea.  I'll play around with that.

milseman:
Do you mean inlined? We can slap an `@inline(__always)` on them.

compnerd:
Yeah, an `@inline(__always)` should do the trick - I was thinking of the clang flatten attribute which basically inlines all the way through a call stack, but the same idea.

Sources/SystemInternals/Syscalls.swift
milseman: Should we make the `FilePermissions` OptionSet empty on Windows? I'm a little uneasy about ignoring information.

compnerd: no, there is the concept of file permissions - its just that the permissions are different, there are no group permissions, and there are no execute permissions - only read/write.

milseman: Ah, then we should we probably have group/execute unavailable on Windows, and the implementation of open here should do the read/write permission setting.

milseman:
Alright, I think https://github.com/apple/swift-system/pull/11 is ready to review. It adds the basic organization and mocking support. If you want attribution for the code, feel free to steal the commits and use them for this branch :-). Up to you. Let me know if I can help with anything else to move this PR forwards.

milseman:
IIUC, the way to resolve the conflict is to remove all changes to `Syscalls.swift`, given that we now have `WindowsSyscallAdapters.swift`

Sources/System/Errno.swift
```diff
@@ -1187,6 +1195,19 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
...
+  /// The scheduled operation was canceled.
+  ///
+  /// The corresponding C error is `ECANCELED`.
+  @_alwaysEmitIntoClient
+  public static var canceled: Errno { Errno(_ECANCELLED) }
+
+  @_alwaysEmitIntoClient
+  @available(*, unavailable, renamed: "canceled")
+  public static var ECANCELED: Errno { canceled }
+#else
```
benrimmington, line 1210: @compnerd If the `internal var _ECANCELLED: CInt { 10103 }` can also be renamed, then this `#if os(Windows)` will exactly match the `#else` below.

Sources/SystemInternals/Syscalls.swift
```diff
@@ -126,10 +128,10 @@ public func system_lseek(
 
 // write
 public func system_write(
-  _ fd: Int32, _ buf: UnsafeRawPointer!, _ nbyte: Int
-) -> Int {
+  _ fd: Int32, _ buf: UnsafeRawPointer!, _ nbyte: CUnsignedInt
+) -> CInt {
```
milseman, line 132: Similar to above.

Sources/System/FileOperations.swift
```diff
@@ -168,7 +168,7 @@ extension FileDescriptor {
     retryOnInterrupt: Bool = true
   ) throws -> Result<Int, Errno> {
     valueOrErrno(retryOnInterrupt: retryOnInterrupt) {
-      system_read(self.rawValue, buffer.baseAddress, buffer.count)
+      Int(system_read(self.rawValue, buffer.baseAddress, numericCast(buffer.count)))
```
milseman, line 171: Ideally, this casting would be done in the body of `read/write` inside the Windows syscall adapters. Can we avoid this at all?

Sources/SystemInternals/Syscalls.swift
```diff
@@ -78,7 +80,7 @@ public func system_open(_ path: UnsafePointer<CChar>, _ oflag: Int32) -> CInt {
 }
 
 public func system_open(
-  _ path: UnsafePointer<CChar>, _ oflag: Int32, _ mode: mode_t
+  _ path: UnsafePointer<CChar>, _ oflag: Int32, _ mode: CInt
```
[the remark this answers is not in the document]
compnerd: Yes, this should be `CModeT` now.

milseman (approved):
LGTM! Thanks for all the cleanup!

PR #8, opened 2020-10-21
Syscall mock testing

milseman:
Syscall mock testing
    
Adds mocking and tracing infrastructure, test
infrastructure, and trace-based tests.


milseman:
@lorentey I think this is ready to merge after your review. Could you make sure I'm doing sane things with Unmanaged?

Tests/SystemTests/TestingInfrastructure.swift
```diff
@@ -0,0 +1,132 @@
...
+        try body(true)
+        self.expectEqual(mocking.trace.dequeue(), self.expected)
+      } catch {
+        self.fail()
+      }
+
+      // Test interupt behavior. Interruptable calls will be told not to
+      // retry to catch the EINTR. Non-interruptable calls will be told to
+      // retry, to make sure they don't spin (e.g. if API changes to include
+      // interruptable)
```
lorentey, line 98: I believe the set of errno cases a particular syscall is allowed to return is pretty much set in stone at this point; API changes are unlikely. (For well-established syscalls at least.)

Package.swift
```diff
@@ -19,7 +19,10 @@ let targets: [PackageDescription.Target] = [
     path: "Sources/System"),
   .target(
     name: "SystemInternals",
-    dependencies: ["CSystem"]),
+    dependencies: ["CSystem"],
+    swiftSettings: [
+      .define("ENABLE_MOCKING")
```
lorentey, line 24: Shouldn't this be conditional on the current build configuration? (Or, if that's even possible, only enabled in a special test target dependency, and never in the production library?)

Sources/SystemInternals/Mocking.swift
```diff
@@ -0,0 +1,164 @@
...
+//
+// NOTE: This is currently the bare minimum needed for System's testing purposes, though we do
+// eventually want to expose some solution to users.
+//
+// Mocking is contextual, accessible through MockingDriver.withMockingEnabled. Mocking
+// state, including whether it is enabled, is stored in thread-local storage. Mocking is only
+// enabled in testing builds of System currently, to minimize runtime overhead of release builds.
+//
+
+public struct Trace {
```
lorentey, line 20: I think the entire mocking infrastructure ought to be compiled out of the codebase when we aren't building for testing -- so this entire file should be inside a `#if ENABLE_MOCKING` block, or it should live in a special target that's only ever imported in the test targets.

lorentey (approved):
I have three concerns, but overall big thumbs up:

- [trivial] The mocking drivers are getting leaked due to a missing release
- [technical objection] The mocking infrastructure should not be present in the binaries downstream packages/apps link with
- [policy/process] A vague worry that this subsystem may become a barrier to contributing to this package.

Sources/SystemInternals/Mocking.swift
```diff
@@ -0,0 +1,164 @@
...
+    defer {
+      if let object = priorMocking {
+        pthread_setspecific(key, Unmanaged.passUnretained(object).toOpaque())
+      } else {
+        pthread_setspecific(key, nil)
+      }
+    }
+
+    let driver = MockingDriver()
+    guard 0 == pthread_setspecific(key, Unmanaged.passRetained(driver).toOpaque()) else {
```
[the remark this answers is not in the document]
milseman: Hmm, I think we should pass it unretained, fix the lifetime, and remove the `releaseObject` destructor above.

Currently, if the thread is killed and the TLS entry is non-NULL, `releaseObject` is called. If this is the top-most `withMockingEnabled`, then the entry will be set to NULL and there will be a leak. If this is a nested call, then the parent driver will be restored (AFAICT this is balanced), and this will leak. If the thread is killed during this call, then the "passRetained" will be paired with the the release in `releaseObject`, IIUC leakiness depends on whether there are other references to the driver alive.

But, if we pass it unretained and fix the lifetime, it will be balanced for the duration of the function and the `defer` above will remove the TLS entry anyways. If we don't have a `releaseObject` registered, then our leaking story if the thread is killed during this function is just the normal one for Swift.

Tests/SystemTests/TestingInfrastructure.swift
milseman: My understanding is that is much more fuzzy. The man pages show one set of errnos, but in practice that is already out of date and can change version-to-version of the OS, as well as unanticipated corner cases involving things like network file systems.

Tests/SystemTests/TestingInfrastructure.swift
```diff
@@ -0,0 +1,132 @@
...
+      do {
+        let oldErrno = mocking.forceErrno
+        mocking.forceErrno = .always(errno: EINTR)
+        defer { mocking.forceErrno = oldErrno }
```
milseman, line 102: Since we changed it make a new driver after `withMockingEnabled`, we don't have to restore this anymore. We can just assert that it is `.none`.

Package.swift
milseman: The best I can do, IIUC, is:
`      .define("ENABLE_MOCKING", .when(configuration: .debug))`

We also need executables as tests to build in debug and release at some point.

Sources/SystemInternals/Mocking.swift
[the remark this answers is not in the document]
lorentey: My point is that the destructor hook isn't called by `pthread_setspecific` -- it is only called at thread exit. When `withMockingEnabled` restores the original driver, it must also release the driver that it replaced. 

(It doesn't seem feasible to prevent all leaks if someone calls `pthread_exit`, but we should make sure not to leak in regular use.)

Slight nit: it would make this code slightly simpler if it simply used `getspecific` instead of `currentMockingDriver` -- it wouldn't need to deal with the back-n-forth conversion and optionals at all.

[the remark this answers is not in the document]
lorentey: Ah, I see what you mean -- removing the destructor could work too!

Sources/SystemInternals/Mocking.swift
lorentey: `@testable` would be a good way to hide these.

But I still think these should be entirely compiled out from build products that external code links with. (I.e., the mock* functions and the if conditions within the syscall shims must not be compiled into production libraries.) 

We are forced to do invasive testing here because we lack good alternatives -- but potentially leaving traces of the testing infrastructure in production code is a hazard I don't like.

milseman: Internal code that is never called can be dropped by the compiler (and even the linker), unless there is something preventing the optimization. I'll do a bit of both.

milseman:
@lorentey I've only enabled mocking in debug builds (best I can do to approximate testing), and I've conditionally compiled out all of the types and most of the infrastructure and use of the infrastructure. The `mock_foo` internal methods are still "present", but they're only called after a constant-foldable false branch and should be eliminated at compilation time.

Package.swift
lorentey: It's a rather unfortunate choice, but we also have the option to build & run tests outside SPM's limitations. (swift-atomics will likely also need to do that to be able to run lit tests for codegen verification.)

In this sort of setup, the test environment would build the package with ENABLE_MOCKING enabled on the swiftpm command line, enabling both the mocking infrastructure and the mocking tests themselves.

Sources/SystemInternals/Mocking.swift
lorentey: I don't think it's a good idea to rely on the optimizer to eliminate these -- the mocking entry points introduce a level of userspace dynamism that feels inappropriate for syscalls.

lorentey:
I don't think it's a good idea to rely on the optimizer to eliminate these -- the mocking entry points introduce a level of userspace dynamism that feels inappropriate for a low-level syscall wrapper. Why not compile these out entirely?

milseman:
I'm normally the poster child for not trusting the optimizer to do much of anything, but trivial dead code elimination is something I'd trust the optimizer to do. The downside of conditionally compiling out the trivial functions would be that the already-onerous bodies of the `system_foo` calls would have even more stuff in them. The only ways I could think of around this involved helper function taking closures, which would be relying on the optimizer to do something less trivial than DCE.

FWIW, I disassembled the .o file and these mocking calls are not present.

milseman:
I think I found a way that does both conditional compilation **and** reduces the boilerplate. It would make it harder to promote this to a user-facing feature in the future, but hopefully we will have migrated to code generation by then.

milseman:
@lorentey Could you give this another pass? I think I improved upon all of your concerns. It is true that this will add yet another concern to contributors, but it's hopefully much less than before and fairly straight forward to follow the pattern. New syscalls would need tests anyways, so there is parity between the `system_foo` and the `MockTestCase(name: "foo"...` code.

lorentey (approved):
This looks great! 🚢

PR #9, opened 2020-10-22
SystemInternals: support mocking on Windows

compnerd:
Add the Windows threading APIs for thread local storage.  Prefer the FLS
allocations over TLS allocations simply because they give us more
similar semantics.

### DIFF
--- a/Sources/SystemInternals/Mocking.swift
+++ b/Sources/SystemInternals/Mocking.swift
@@ -87,10 +87,22 @@ public class MockingDriver {
 import Darwin
 #elseif os(Linux) || os(FreeBSD) || os(Android)
 import Glibc
+#elseif os(Windows)
+import ucrt
+import WinSDK
 #else
 #error("Unsupported Platform")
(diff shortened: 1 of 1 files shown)


compnerd:
@milseman - let me know if you have an idea of structuring the change here better.  I suppose that we can create aliases for `ThreadLocalSet` and `ThreadLocalGet` for the cases where the return value is ignored, but the return conditions are different, which makes it a bit trickier.

milseman:
Alright, I'll try to do some refactoring. Thanks for this!

milseman:
https://github.com/apple/swift-system/pull/10

milseman:
In https://github.com/apple/swift-system/pull/11, I actually implement this (I believe). Would you be open to me merging that PR after I slap on `@inline(__always)`, closing this PR, and you following up with more Windows support on top of that?

milseman:
I think this can be closed. Feel free to reopen if I'm mistaken

PR #11, opened 2020-10-23
Windows syscall adapters

milseman:
Set up some code organization to pave the way for @compnerd's awesome Windows support work.

PR #12, opened 2020-10-23
Follow up to #11 to repair the build

compnerd:


Sources/SystemInternals/WindowsSyscallAdapters.swift
```diff
@@ -35,6 +38,7 @@ internal func close(_ fd: Int32) -> Int32 {
 internal func lseek(
   _ fd: Int32, _ off: off_t, _ whence: Int32
 ) -> off_t {
+  // FIXME(compnerd) should this use `_lseeki64` to enable large file access?
```
milseman, line 41: On Darwin, the `off_t` is a 64bit int. Is Windows different?

Sources/SystemInternals/WindowsSyscallAdapters.swift
```diff
@@ -9,16 +9,19 @@
...
+internal func open(_ path: UnsafePointer<CChar>, _ oflag: Int32) -> CInt {
   var fh: CInt = -1
   _ = _sopen_s(&fh, path, oflag, _SH_DENYNO, _S_IREAD | _S_IWRITE)
   return fh
 }
 
 @inline(__always)
 internal func open(
-  _ path: UnsafePointer<CChar>, _ oflag: Int32, _ mode: mode_t
+  _ path: UnsafePointer<CChar>, _ oflag: Int32, _ mode: CInt
```
milseman, line 24: I think we should keep the `mode_t` in this interface and do the numeric cast in the body. If Windows doesn't have `mode_t`, then we can have a public typealias for `CModeT` in `Exports.swift`. I was phrasing the `system_foo` functions as using the non-typealiased decls from Darwin/GlibC, but we can change that too if it makes sense.

Sources/SystemInternals/WindowsSyscallAdapters.swift
```diff
@@ -9,16 +9,19 @@
 
 #if os(Windows)
 
+import ucrt
+import WinSDK
+
 @inline(__always)
-internal func open(_ path: UnsafePointer<CChar>, _ oflag: Int32) {
+internal func open(_ path: UnsafePointer<CChar>, _ oflag: Int32) -> CInt {
```
milseman, line 16: Whoops, good catch.

milseman (approved):
Until we can get CI, is there a good way for me to make sure changes don't break building on Windows?

compnerd:
It should be possible to build on windows locally using the snapshots.  You could grab a vm image from Microsoft at https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/ for development purposes.

Sources/SystemInternals/WindowsSyscallAdapters.swift
compnerd: Yeah, windows and Linux both do separation of 32-bit and 64-bit.  The off_t is 32-bit.

Sources/SystemInternals/WindowsSyscallAdapters.swift
compnerd: Oh, if we're introducing a CModeT type alias, that's perfect.

Sources/SystemInternals/WindowsSyscallAdapters.swift
milseman: I'm in favor of always using the 64-bit variation

milseman (approved):
LGTM

PR #14, opened 2020-11-05
FilePath syntactic operations

milseman:
Syntactic operations for FilePath, inspired by Rust and C++17. Includes Windows support.

[Review thread](https://forums.swift.org/t/api-review-filepath-syntactic-apis-version-2/44197)


Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+  ///   * `/`
+  /// * Windows:
+  ///   * `C:\Users\`
+  ///   * `\\?\UNC\server\share\bar.exe`
+  ///   * `\\server\share\bar.exe`
+  ///
+  /// TODO: We should add API to system to access environment variables
+  /// and perform expansion/substituion, but that's deferred for now. Should
+  /// we mention that?
+  public var isAbsolute: Bool {
```
abertelrud, line 30: I still think that `RelativeFilePath` and `AbsoluteFilePath` (possibly shortened to just `FilePath`) should be primary types, so that API can declare their assumptions at compile-time rather than throw runtime errors if the "wrong" kind of path is passed.  I understand that this isn't how most APIs represent path types, but I think it's worked quite well in ToolSupportCore and SwiftPM.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+  ///
+  /// Examples:
+  /// * Unix:
+  ///   * `~/bar`
+  ///   * `tmp/foo.txt`
+  /// * Windows:
+  ///   * `bar\baz`
+  ///   * `C:Users\`
+  ///   * `\Users`
+  public var isRelative: Bool { !isAbsolute }
```
abertelrud, line 44: I'm not sure this is needed, since it's unambiguously defined as "not absolute".

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+
+  /// A textual representation of the file path using the generic separator (`/`).
+  ///
+  /// On Unix, this is equivalent to `description`
+  ///
+  /// TODO: Should `dump()` use this? Also, should dump escape invalid Unicode?
+  public var genericDescription: String {
```
abertelrud, line 51: Is this intended only for display purposes?  If so, perhaps it should be called `displayDescription`?  Is there also a way to get a string form that is guaranteed to be reversible, for example by serializing to a JSON and being able to be read back in?

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+    fatalError()
+  }
+
+  /// Returns whether `self` is a prefix of `other`
+  public func starts(with other: FilePath) -> Bool {
```
abertelrud, line 56: I might be reading this wrong, but is the description backwards?  If self starts with other, then other is a prefix of self.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+    fatalError()
+  }
+
+  /// Returns whether `self` is a suffix of `other`
+  public func ends(with other: FilePath) -> Bool {
```
abertelrud, line 61: I might be reading this wrong, but is the description backwards?  If self ends with other, then other is a suffix of self.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+    fatalError()
+  }
+
+  /// Whether this path is empty
+  public var isEmpty: Bool { self.storage == [0] }
```
abertelrud, line 66: I've never found this a useful question to ask, but perhaps it's needed.  If this is really rare, could clients instead do `myPath == ""` (since Strings are convertible to FilePaths)?

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+  /// haven't run tests). Rust differs in that has_root is just looking for the `\` somewhere.
+  ///
+  /// REVIEW NOTE: Going with UNC or device paths not including trailing separator,
+  /// which is consistent with C# IIUC. I need to perform the experiement to verify this.
+  ///
+  /// REVIEW NOTE: I'm going with component as return type, which implies the more common
+  /// use would be to read or inspect the root, rather than forming new paths from the same root.
+  /// For that use case, you can do `FilePath(otherPath.root)`, or use
+  /// the `relativePath`'s setter, or use the `/` operator (if we go that route).
+  public var root: FilePath.Component? {
```
abertelrud, line 114: This makes a lot of sense to be based on my understanding of Windows paths.

abertelrud: If we _did_ have `AbsolutePath` and `RelativePath`, this is the kind of property that would make sense to have as non-optional for `AbsolutePath` and that just wouldn't be on the `RelativePath` type at all.  Makes things a bit cleaner for the client when they know they have an absolute path, for example (because that's the contract and it has been asserted) and then they can just access the root without having to unwrap.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+// Decompose a path into its parts
+//
+//   path = root? relative-path?
+//   root = root-name? root-directory?
+//   root-directory = separator
+//   root-name (Windows only) = disk | UNC | device
+extension FilePath {
+  /// Returns the root directory of a path if there is one, otherwise `nil`
+  ///
+  /// On Unix, this will return the leading `/` if the path is absolute
```
abertelrud, line 79: I think this makes a lot of sense.  For TSCs/SwiftPM's `AbsolutePath` we debated whether the parent directory of the root path should be nil, meaning that every directory might potentially not have a parent, but that gets really unergonomic at the call sites.  Framing this in terms of what the root is, and having the root have itself as the root, makes a ton of sense I think.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+  public var relativePath: FilePath {
+    get {
+      let (_, relIdx) = _parseRoot()
+      return FilePath(slice: storage[relIdx...])
+    }
+    set { fatalError() }
+  }
+
+  /// Returns the final component of the path. If the path is a normal file, this
+  /// will be the file name. If the path is a directory, this is the directory name.
```
abertelrud, line 154: I'm not sure that the mention of a file vs a directory makes sense here, since this is just involved with file path syntax.  Seems to me that at this level of the API, the discussion is just around a sequence of path components, whether or not they even exist in the file system, or whether they are symlinks, special devices, or other things.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+  /// Returns `nil` if the path is empty or only contains a root
+  ///
+  /// REVIEW NOTE: We follow  `basename` semantics, unlike `file_name` in
+  /// C++ which treats trailing `/` specially, or in Rust which returns nil for `..`
+  ///
+  /// REVIEW NOTE: shell and C basename will return `/` if the path is `/`. I don't
+  /// think we want to, it seems clearer that this only is in the relative portion of the path,
+  /// and doing so would make the setter weird.
+  public var basename: FilePath.Component? {
```
abertelrud, line 163: I understand the reasoning for making this optional, but I think it makes the API fairly unergonomic for clients.  Pretty much every client that wasn't to know the last path component would need to check for it and throw an error.  Maybe that's the correct behavior, but it seems a bit tedious.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+      _ = self.pop()
+      if let c = newValue { self.append(FilePath(c)) }
+    }
+  }
+
+  /// Returns the path up to but not including the `basename`.
+  ///
+  /// If the path only contains a root, returns `self`.
+  /// If the path has no root and only includes a single component,
+  /// returns an empty FilePath.
```
abertelrud, line 182: This seems like the right choice to me, but then I would expect `basename` to also return an empty string rather than optional.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  /// The extension of the file or directoy last component.
+  ///
+  /// If `basename` is `nil` or one of the special path components
+  /// `.` or `..`, `get` returns `nil` and `set` does nothing.
+  ///
+  /// If `basename` does not contain a `.` anywhere, or only
+  /// at the start, `get` returns `nil` and `set` will append a
+  /// `.` and `newValue` to `basename`
```
abertelrud, line 201: I think there are additional rules on some platforms, such as not treating extensions consisting only of digits as extensions (e.g. "Foo 1.0" doesn't have the extension "0"), but I'm guessing that might be a bit too specialized for this API.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+  ///   * `/tmp/foo.txt => "foo"`
+  ///   * `/Appliations/Foo.app/ => Foo`
+  ///   * `/Appliations/Foo.app/bar.txt => bar`
+  ///   * `/tmp/.hidden => .hidden``
+  ///   * `/tmp/.. => ..``
+  ///   * `/ => nil`
+  ///
+  /// TODO: setter? That's a little harder to imagine than for `extension`.
+  /// Setting to `nil` is more likely a bug.
+  public var stem: String? {
```
abertelrud, line 235: Since it's only the root that doesn't have a stem, it feels a little odd to have this be optional, because of the effect on the callsites.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+
+  /// Reserve enough storage space to store `minimumCapacity` path characters
+  public mutating func reserveCapacity(_ minimumCapacity: Int) {
+    self.storage.reserveCapacity(1 + minimumCapacity)
+  }
+
+  /// Append the contents of `other`.
+  ///
+  /// TODO: What if `other` is absolute? C++ does replacement for append
+  /// but not for concat... Rust's `push` does replacement. Should we just do push/pop?
```
abertelrud, line 259: This is where I would again argue for separate `AbsolutePath` and `RelativePath` types.  :)  It makes the intent clear, not just here, but also for any client APIs that take or return paths.  There are a few edge cases where something could be absolute or relative, but in many of those cases it's appropriate to use an enum anyway, at least in the cases where this has come up in SwiftPM.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+    self.components.append(contentsOf: other.components)
+  }
+
+  /// Remove and return the last component of this file path. If the path is
+  /// root or empty, does nothing and returns `nil`.
+  ///
+  /// Examples:
+  /// * `"/".pop() == nil          // path is /`
+  /// * `"/foo/bar".pop() == "bar" // path is /foo`
+  public mutating func pop() -> FilePath.Component? {
```
abertelrud, line 272: I would suggest `popLast` here to be a bit more clear (not a big fan of the word "pop" for data structures that aren't specifically stacks, but there's certainly precedent for that in Swift).

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+    self.components.popLast() // FIXME: popLastButNotRoot()
+  }
+
+  /// See `append`
+  ///
+  /// TODO: Should this be `/`? Should the RHS have to be Component?
+  public static func +(_ lhs: FilePath, _ rhs: FilePath) -> FilePath {
```
abertelrud, line 279: I think `+` is the right choice here.  One concern with the operator combined with the string literal conversion is that it isn't obvious whether each `foo + "usr" + "local" + "include"` does, based on whether `foo` is a path or string; in that case it looks a lot like string concatenation.  But `/` doesn't feel right since it's specific to the Unix form of the path.

If the RHS has to be a component, then there would need to be some other way to append a path.

In TSC's `AbsolutePath` and `RelativePath` we avoided the `+` operator for this reason.  Instead there is `AbsolutePath.append(_ subpath: RelativePath)` and `AbsolutePath.append(component: String)` (as well as `AbsolutePath.append(components: String...)`), but that does get a bit cumbersome to type.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,353 @@
...
+    /*public*/ var kind: Kind { fatalError() }
+    /*public*/ var isVerbatim: Bool { fatalError() }
+  }
+  /*
+   DOS: letter ':'
+   UNC: '\\' host name '\' (share name | letter '$')
+   DOS device:
+   ('\\.\' | '\\?\') (letter ':' | 'Volume{' GUID '}')
+   */
+  /*public*/ var windowsRootName: WindowsRootName? { nil }
```
abertelrud, line 350: Will there also be a way to get at the various roots on Windows, e.g. a static function to return the root for a particular drive letter on Windows?  Or would that always be done by just creating a particular path with a `C:\` string passed in?

abertelrud:
Overall this looks great!  Really excited to see this API!

Sources/System/FilePath/FilePathSyntax.swift
milseman: A path on Windows can be printed in a platform-preferred representation using backslash, or in a more generic/portable fashion using forwards slashes. `FilePath.description` uses the backslash. We do need a good name for this and I was also considering "portableDescription".

We might make this only available on Windows since its presence is unneeded cognitive overhead for POSIXy users. The counter argument would be that code which wants this representation would need the `#if` in it.

Sources/System/FilePath/FilePathSyntax.swift
milseman: I get what you're saying. I think it's a useful aid to reduce cognitive burden because "relative" is a concept that is used elsewhere, even though it is trivially derivable.

Sources/System/FilePath/FilePathSyntax.swift
milseman: We could consider adding a `FilePath` protocol and concrete types that conform and precondition on construction. This can be added at any time and I'd prefer to defer that until after we add semantic operations, since I think a lot of the relative-vs-absolute concerns come up then.

IIRC, SPM doesn't even use the same semantics for absolute paths as we do here, so you might not get what you want if System provides these types. This would argue to defer this until we have access to the environment for `~` expansion, and even `realpath` functionality.

Sources/System/FilePath/FilePathSyntax.swift
milseman: I did just do a push that implements this, adds tests, and fixes the docs. Now it is whole-component-wise prefix/suffix matching.

Sources/System/FilePath/FilePathSyntax.swift
milseman: Yup, same

Sources/System/FilePath/FilePathSyntax.swift
milseman: True, though that might trigger a new allocation (this implementation might as well). This is consistent with other types, such as Collections, which have a concept of emptiness. It is true that anything with a empty initializer (or empty-string initializer) and implements Equatable doesn't need this method, but it's useful in more generic contexts.

FWIW, in implementing algorithms over the backing storage for FilePath, I use `isEmpty` a lot over comparing to a literal, especially since that's useful even if I change the backing representation.

Sources/System/FilePath/FilePathSyntax.swift
milseman: Relative paths on Windows can have roots: `C:` and `\`.

Sources/System/FilePath/FilePathSyntax.swift
milseman: I was trying to communicate that this was a syntactic consideration and doesn't ask the filesystem whether this is a file or directory, but the wording might be confusing. I'm going with something like this now:

```swift
  /// Returns the final relative component of the path.
  /// Returns `nil` if the path is empty or only contains a root
  ///
  /// Examples:
  /// * Unix:
  ///   * `/usr/local/bin/ => bin`
  ///   * `/tmp/foo.txt    => foo.txt`
  ///   * `/               => nil`
  /// * Windows:
  ///   * `C:\Users\                    => Users`
  ///   * `C:Users\                     => Users`
  ///   * `C:\                          => nil`
  ///   * `\Users\                      => Users`
  ///   * `\\?\UNC\server\share\bar.exe => bar.exe`
  ///   * `\\server\share               => nil`
  ///   * `\\?\UNC\server\share\        => nil`
  ///

```

Sources/System/FilePath/FilePathSyntax.swift
milseman: It's always a debate whether presence or non-presence of a result should be elevated to the type system rather than a convention on a particular return value. If it's salient we like to promote it to the type level. Paths can be empty, but components generally cannot (at least if they were extracted from a path). The empty path has no components, rather than it having a single empty component. Inserting an empty component is a weird thing to do and indicative of a potential bug, so we try to avoid empty components. We may even want to forbid empty components.

`relativePath` up above and `dirname` below uses empty paths because every path has a `dirname` and `relativePath`, which could be empty. `root` is optional not only because the presence or non-presence is highly meaningful, but also because it is a component. These component APIs also have `setters`, where it is useful to set a component to `nil`, which I feel is clearer algebraically than `""`.

Sources/System/FilePath/FilePathSyntax.swift
milseman: That's an interesting point. Would this be a platform-specific consideration or a file-system-specific consideration or application-specific?

Sources/System/FilePath/FilePathSyntax.swift
milseman: Windows can use the `root` member to get the root out, which would be the whole root (ala `C#`'s semantics). I do want to provide further root analysis, akin to what Rust's [Prefix](https://doc.rust-lang.org/std/path/enum.Prefix.html) type can do, though I'm more inclined to go with naming and structure that would be familiar to Windows users over Rust's naming. I have some corner cases to figure out there as well, so I need to run some experiments.

FWIW, what is currently proposed is as much as C++ or C# gives users, I just would like to give even more (especially since I have to write the root parser anyways).

Sources/System/FilePath/FilePathSyntax.swift
milseman: I think if we do an operator it will be `/` instead of `+`, since we won't be modeling "concatenation" in a stringy sense. C++ uses `/` and it seems to be clear and natural even on Windows (where `/` can be used as a path separator prior to normalization). Not using an operator is also fine.

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: Understood.  I do think that `~` expansion and `realpath` would probably be outside of a `FilePath` value type (especially `realpath`, since it depends on which file system is involved), but that would be a later discussion as well.

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: That makes sense.

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: Right, I do think that we'd want to avoid `#if` at the callsites where possible.  Maybe a better way for me to phrase the question is "what API would a client use to get a portable description that can be stored in a file and reconstituted on any platform".

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: This makes sense.  Thanks for explaining.

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: Ah, right, that's true.  I forgot about that, and that certainly complicates things.

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: Thanks, that's much clearer.

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: It would be a platform-specific consideration.  I looked to see if I could find anything more specific about macOS, which is the platform I was thinking of with the no-digits consideration, but I couldn't find anything relevant so I may be misremembering.  This was something that was discussed a lot around the time of Mac OS X 10.0, when the Classic MacOS semantics were being reconciled with the Unix-based conventions of OPENSTEP.  But I think I'm either misremembering or things have changed, since it's just that the `.` can't be the first or last character of the last path component.

Sources/System/FilePath/FilePathSyntax.swift
abertelrud: What I meant by my question was whether on Windows, I could create a path that had the root corresponding to a particular drive letter (such as one entered by a user) without already having a `FilePath` instance from which to get a root.

Sources/System/FilePath/FilePathSyntax.swift
milseman: Resolution: Defer this property from this pitch. Ideally we'd look at RFC-1808 for guidance. Corner cases are application specific. String is not the ideal round trip format due to encoding correction.

Sources/System/FilePath/FilePathString.swift
```diff
@@ -0,0 +1,251 @@
...
+
+// @available(macOS 10.16, iOS 14.0, watchOS 7.0, tvOS 14.0, *)
+extension String {
+  /// Creates a string by interpreting the file path's content as UTF-8 on Unix
+  /// and UTF-16 on Windows.
+  ///
+  /// - Parameter path: The file path to be interpreted as `PlatformUnicodeEncoding`.
+  ///
+  /// If the content of the file path isn't a well-formed Unicode string,
+  /// this initializer replaces invalid bytes them with U+FFFD.
```
AverageHelper, line 97: We probably shouldn't leave "invalid bytes them" in the documentation comments. Should this sentence refer to "invalid bytes" or to "them"?

This phrase occurs in a few other places here as well.

milseman: Good catch. Dropping "them".

Sources/System/FilePath/FilePathString.swift
```diff
@@ -0,0 +1,251 @@
...
+  /// this replaces invalid bytes them with U+FFFD. See `String.init(decoding:)`
+  public var debugDescription: String { description.debugDescription }
+}
+
+// @available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
+extension String {
+  /// Creates a string by interpreting the path component's content as UTF-8 on Unix
+  /// and UTF-16 on Windows.
+  ///
+  /// - Parameter path: The path component to be interpreted as `PlatformUnicodeEncoding`.
```
AverageHelper, line 144: The initializer methods in this extension appear to be documented incorrectly. Please correct me if I am wrong, but I believe IDEs would expect `- Parameter path:` to describe a parameter that is not named in any of these signatures. Should these documentation comments instead read `- Parameter component`?

Also, below is another instance of "invalid bytes them".

somy55:
J

Sources/System/FilePath/FilePathString.swift
somy55: J

Sources/System/FilePath/FilePathComponentView.swift
```diff
@@ -0,0 +1,220 @@
...
+        self = comp._path
+
+        if !rootStr.isEmpty {
+          if let r = self.root {
+            // Roots can be forgotten but never altered
+            assert(r._slice.elementsEqual(rootStr))
+          }
+          // TODO: conditional on it having changed?
+          self.root = Root(rootStr)
+        }
```
kylemacomber, line 65: I think there's a small issue here, where a path with an empty root can get its root stomped:

```swift
let relative: FilePath = "usr/local/bin"
let absolute: FilePath = "/usr/local/bin"

var path = relative
path.components = absolute.components
XCTAssertEqual(path, relative)
```

```
XCTAssertEqual failed: ("/usr/local/bin") is not equal to ("usr/local/bin")
```

Sources/System/FilePath/FilePathComponentView.swift
```diff
@@ -0,0 +1,220 @@
...
+      self._start = path._relativeStart
+      _invariantCheck()
+    }
+  }
+
+  /// View the non-root components that make up this path.
+  public var components: ComponentView {
+    __consuming get { ComponentView(self) }
+    _modify {
+      // RRC's empty init means that we cann't guarantee that the yielded
```
kylemacomber, line 46:
```suggestion
      // RRC's empty init means that we can't guarantee that the yielded
```

Sources/System/FilePath/FilePathComponentView.swift
```diff
@@ -0,0 +1,220 @@
...
+    return Index(_path._parseComponent(startingAt: i._storage).nextStart)
+  }
+
+  public func index(before i: Index) -> Index {
+    Index(_path._parseComponent(priorTo: i._storage).lowerBound)
+  }
+
+  public subscript(position: Index) -> FilePath.Component {
+    let end = _path._parseComponent(startingAt: position._storage).componentEnd
+    return FilePath.Component(_path, position._storage ..< end)
```
kylemacomber, line 102: Is it worth making ComponentView.Index store a Range so subscript is O(1)?

milseman: It is amortized constant, as in a subscript over every index will be a total of O(n). We could store the range, but that would complicate `startIndex` and/or the initializer for ComponentView. If we have an `unknown` representation for the length, that further complicates other logic. `subscript` is not the only operation to optimize for, as many operations (especially RRC operations) only advance indices or slice the collection.

This "parse" is just scanning bytes looking for the `/` byte (or `\` on Windows). I don't think it's worth spending the complexity budget on, since these types are resilient, these operations are not inlinable, and the scanned byte range is loaded at the end (with a retain, at least until we do small-form for small components).

kylemacomber: It's good we have the flexibility to add this in the future.

Computing the Range in the ComponentView initializer seems like the right direction to me. It's the direction we've taken with the [`ChunkedByCount`](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Chunked.swift#L285-L290)  and [`SlidingWindows`](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/SlidingWindows.swift#L45) collections from the Algorithms package.

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,616 @@
...
+  /// at the start, `get` returns `nil` and `set` will append a
+  /// `.` and `newValue` to `lastComponent`.
+  ///
+  /// Otherwise `get` returns everything after the last `.` and `set` will
+  /// replace the extension.
+  ///
+  /// Examples:
+  ///   * `/tmp/foo.txt                 => txt`
+  ///   * `/Appliations/Foo.app/        => app`
+  ///   * `/Appliations/Foo.app/bar.txt => txt`
```
kylemacomber, line 310:
```suggestion
  ///   * `/Applications/Foo.app/        => app`
  ///   * `/Applications/Foo.app/bar.txt => txt`
```

kylemacomber: You'll need to re-indent the rest of this comment to get the => arrows to line up

Sources/System/FilePath/FilePathSyntax.swift
```diff
@@ -0,0 +1,616 @@
...
+  /// * `../usr/local/../bin => ../usr/bin`
+  public mutating func lexicallyNormalize() {
+    defer { _invariantCheck() }
+    _normalizeSpecialDirectories()
+  }
+
+  /// Returns a copy of `self` in lexical-normal form, that is `.` and `..`
+  /// components have been collapsed lexically (i.e. without following
+  /// symlinks). See `lexicallyNormalize`
+  public var lexicallyNormal: FilePath {
```
kylemacomber, line 394: It's a little surprising to me that this is a computed property `lexicallyNormal` rather than a method `lexicallyNormalized()` because it has a mutating counterpart `lexicallyNormalize()`. (I think `isLexicallyNormal` is fine).

kylemacomber (approved):
Looks great! I only had minor additional comments (and didn't have time to carefully read through the tests).

Sources/System/UtilConsumers.swift
```diff
@@ -0,0 +1,73 @@
...
+    return _eat(count: es.count)
+  }
+
+  internal mutating func _eatUntil(_ idx: Index) -> Slice {
+    precondition(idx >= startIndex && idx <= endIndex)
+    defer { self = self[idx...] }
+    return self[..<idx]
+  }
+
+  internal mutating func _eatThrough(_ idx: Index) -> Slice {
```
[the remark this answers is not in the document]
milseman: They could be here. It depends on what the total overload set is, as argument labels don't play as well with trailing closures. Either way, these are temporary helpers that I can hopefully drop in the future.

Sources/System/FilePath/FilePathParsing.swift
```diff
@@ -0,0 +1,395 @@
...
+      let wasSeparator = isSeparator(self[readIdx])
+      self.swapAt(writeIdx, readIdx)
+      self.formIndex(after: &writeIdx)
+      self.formIndex(after: &readIdx)
+
+      while wasSeparator, readIdx < endIndex, isSeparator(self[readIdx]) {
+        self.formIndex(after: &readIdx)
+      }
+    }
+    self.removeLast(self.distance(from: writeIdx, to: readIdx))
```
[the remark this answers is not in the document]
milseman: Now that we have a prepass to convert `/` to `\` on Windows, it is. It used to also do the separator change here, but that makes root parsing significantly more complex.

Sources/System/SystemString.swift
```diff
@@ -0,0 +1,289 @@
...
+//
+// Always null-terminated.
+internal struct SystemString {
+  internal typealias Storage = [SystemChar]
+  internal var nullTerminatedStorage: Storage
+}
+
+extension SystemString {
+  internal init() {
+    self.nullTerminatedStorage = [.null]
```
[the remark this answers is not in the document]
milseman: It can be, but IIRC how you declare a global matters and we don't want to unnecessarily dirty memory. I never understood all the differences (especially as they interact with `@usableFromInline`, which wouldn't be relevant here yet).

Sources/System/FilePath/FilePathComponentView.swift
milseman: By "resilient" I wasn't referring to the ability to add it later (though that's true), I was referring to the performance characteristics of the code. Inside the call, subscript does something computers are really good at (linear scan within a cache line for data that will be loaded anyways), after having been called indirectly and returning its result indirectly.

From a performance perspective, I don't think `ChunkedByCount` and `SlidingWindows` are analogous, as their collection status is primarily for the purpose of yielding elements. They're not RRC, for example.

PR #15, opened 2020-11-14
Code cleanup; FilePath separator normalization

milseman:
Splitting off some of the work being done in https://github.com/apple/swift-system/pull/14. This brings in the separator normalizing behavior as well as re-arranging some of the source code.


Sources/System/FilePath/FilePathParsing.swift
```diff
@@ -0,0 +1,112 @@
...
+
+// The separator we use internally
+private var genericSeparator: SystemChar {
+  numericCast(UInt8(ascii: "/"))
+}
+
+// The platform preferred separator
+//
+// TODO: Make private
+private var platformSeparator: SystemChar {
```
kylemacomber, line 20: Since this is private I think you can remove the TODO.

Sources/System/FilePath/FilePathParsing.swift
```diff
@@ -0,0 +1,112 @@
...
+#if os(Windows)
+  return numericCast(UInt8(ascii: "\\"))
+#else
+  return genericSeparator
+#endif
+}
+
+// TODO: Make private
```
kylemacomber, line 28:
```suggestion
```

Sources/System/FilePath/FilePathParsing.swift
```diff
@@ -0,0 +1,112 @@
...
+private func isSeparator(_ c: SystemChar) -> Bool {
+  c == genericSeparator || c == platformSeparator
+}
+
+// @available(macOS 10.16, iOS 14.0, watchOS 7.0, tvOS 14.0, *)
```
kylemacomber, line 33:
```suggestion
```

kylemacomber: Since none of this is public I don't think we need (commented out) availability? Since this is new code, I'm not sure this availability annotation would be accurate?

Sources/System/FilePath/FilePathParsing.swift
```diff
@@ -0,0 +1,112 @@
...
+extension FilePath {
+  // For invariant enforcing/checking. Should always return `nil` on
+  // a fully-formed path
+  private func _trailingSepIdx() -> Storage.Index? {
```
kylemacomber, line 37:
```suggestion
  private func _trailingSeparatorIndex() -> Storage.Index? {
```

The abbreviation here seems ad hoc.

Sources/System/FilePath/FilePathParsing.swift
```diff
@@ -0,0 +1,112 @@
...
+  // Precondition: There is exactly zero or one trailing slashes
+  //
+  // Postcondition: Path is root, or has no trailing separator
+  private mutating func _removeTrailingSeparator() {
+    assert(bytes.last == 0, "even empty paths have null termiantor")
+    defer { assert(_trailingSepIdx() == nil) }
+
+    guard let sepIdx = _trailingSepIdx() else { return }
+    bytes.remove(at: sepIdx)
+
```
kylemacomber, line 54:
```suggestion
```

kylemacomber (approved):
Looks good! I recall seeing some unit tests that looked like they exercised the separator normalization logic. Are those already on main? If not, should we pull them in as part of this patch?

Sources/System/FilePath/FilePathParsing.swift
milseman: It is necessary due to a compiler limitation where non-public code that uses a public type needs availability. But we can defer adding that until it's relevant to an ABI release

milseman:
Closing this, FIlePath syntactic operations has diverged significantly. I could consider doing some of the refactoring in a separate PR, though it might make more sense to just do it as a rebasing on that PR.

PR #22, opened 2021-02-06
Autogenerate platform constants

lorentey:
This PR replaces the manually maintained platform constant definitions with source code generation, simplifying package maintenance.

This is done by adding a new Swift script that looks up a simple (platform-specific) list of symbolic constants in the current platform's SDK, and emits a corresponding `<Platform>PlatformConstants.swift` file in the `autogenerated/` subdirectory of the package sources, as well as a `<Platform>PlatformConstantsTests` test case in the package tests that verifies that the generated values are matching the contents of the SDK used to build the package.

Because the Swift Package Manager has no support for source generation, these files need to be manually committed into the repository.

The script needs to be run every time constants are added or removed, or in case the generated test case reports a failure.
(This latter case will typically require some additional investigation.)


lorentey:
This is currently only tested on Darwin.

milseman (commented):
I like this direction! I wonder if we should make the generator an executable target. It would depend on System+SystemInternals+CSystem modules rather than Foundation+Darwin. Then we can add a test target that runs the executable with a different output file and diff the results to what is checked in.

This would give us self-hosting of a non-trivial script, automatic regression tests, and since we check the result in anyways, we will be bootstrapped.

Future directions here could be the ability to declare types in addition to constants.

lorentey:
Closing -- this whole area is obsoleted by https://github.com/apple/swift-system/pull/29. (We can reopen if we find a reason not to merge that.)

PR #25, opened 2021-02-09
Ease binary distribution

milseman:
Changes to make it easier to work with binary distributions of System.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,7 +29,10 @@ let targets: [PackageDescription.Target] = [
 
   .testTarget(
     name: "SystemTests",
-    dependencies: ["SystemPackage"]),
+    dependencies: ["SystemPackage"],
+    swiftSettings: [
+      .define("SYSTEM_PACKAGE")
+    ]),
 ]
 
 let package = Package(
--- a/Package.swift
+++ b/Package.swift
@@ -15,18 +15,14 @@ import PackageDescription
 let targets: [PackageDescription.Target] = [
   .target(
     name: "SystemPackage",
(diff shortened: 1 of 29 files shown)


Sources/CSystem/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module CSystem {
```
weissi, line 1: I think this will cause issues with SwiftPM. Basically SwiftPM has a [magic codepath](https://github.com/apple/swift-package-manager/blob/main/Sources/Xcodeproj/pbxproj.swift#L587-L596) which makes generated Xcode projects with _if and only if_ there is an umbrella header called `ExactModuleName.h`, ie. `CSystem.h` in this case.

If the name of the umbrella header either diverges or there's a custom module map, then the generated Xcode projects will be much less compatible (don't work as sub-projects) as I described in this issue: https://bugs.swift.org/browse/SR-9024 (that was before I knew that we need to hit this magic codepath).

lorentey: AFAIK, SPM's Xcode project generation feature is de facto obsoleted by Xcode's direct support for packages these days. (I don't think it's actively maintained at this point; trying to support its quirks seems like a bad idea at this point.) 

Omitting the modulemap is an easy enough fix for now, but if we ever need to set special module attributes, then that'll put an end to even this minimal support for `generate-xcodeproj`.

milseman: I'll leave it out for now. In the future we can make a `CSystem.h` header if we need it

PR #21, opened 2021-02-06
Add symbolic names for standard file descriptors

lorentey:
These let us easily read/write data to/from stdin/stdout/stderr without resorting to `FileDescriptor(rawValue: 1)`.

```swift
var greeting = "Hello, world!\n"
try greeting.withUTF8 { buffer in
  _ = try FileDescriptor.standardOutput.writeAll(buffer)
}
```

PR #28, opened 2021-02-24
Add support for dup/dup2

lorentey:
This adds support for duplicating file descriptors via a `FileDescriptor.duplicate` method:

```swift
extension FileDescriptor {
  public func duplicate(
      as target: FileDescriptor? = nil,
      retryOnInterrupt: Bool = true
    ) throws -> FileDescriptor
}
```

The method dispatches to either `dup` or `dup2` based on whether or not it has an explicit target.

Usually, a non-nil `target` will be one of the file descriptors conventionally reserved for the stdin/stdout/stderr role; these were previously assigned symbolic names in https://github.com/apple/swift-system/pull/21. To target other file descriptors, users will need to use the `.init(rawValue:)` initializer. (This will rarely be needed, but some like to pass additional communication channels down to child processes using some custom numbering convention beyond the universal 0/1/2.)

```swift
let controlFd = try fd.duplicate(as: FileDescriptor(rawValue: 3))
```

### DIFF
--- a/Sources/System/FileOperations.swift
+++ b/Sources/System/FileOperations.swift
@@ -309,4 +309,66 @@ extension FileDescriptor {
       buffer,
       retryOnInterrupt: retryOnInterrupt)
   }
+
+  /// Duplicate this file descriptor and return the newly created copy.
+  ///
+  /// - Parameters:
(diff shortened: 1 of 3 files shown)


milseman (approved):
LGTM!

PR #29, opened 2021-02-26
Normal imports

milseman:
System is not an overlay and will not pull C and platform modules into the
user's namespace. Import constants normally rather than redefining them.


Sources/System/Internals/Constants.swift
```diff
@@ -0,0 +1,471 @@
...
+#endif
+
+// MARK: errno
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+@_alwaysEmitIntoClient
+internal var _ERRNO_NOT_USED: CInt { 0 }
+#endif
+
+@_alwaysEmitIntoClient
+internal var _EPERM: CInt { EPERM }
```
lorentey, line 32: I think we should just get rid of these. Is there a good reason to keep them?

Sources/System/Internals/Exports.swift
```diff
@@ -7,19 +7,19 @@
  See https://swift.org/LICENSE.txt for license information
 */
 
-@_implementationOnly import CSystem
+import CSystem
```
lorentey, line 10: We'll need to keep the `CSystem` import `@_implementationOnly`, unless we're willing to make it public. (Which I don't think we would want to do in the ABI-stable incarnation of System.)

lorentey (approved):
Nice! Comments/questions below.

Sources/System/Internals/Constants.swift
milseman: They help avoid the `#if/#elseif/#else` import dance everywhere, and (I didn't check the actual impact of this) they might help when there are other `@unavailable(renamed:)` members with the same name nearby.

Sources/System/Internals/Exports.swift
milseman: Err, so how would that work? Linux gets it definitions from `CSystem`, not `GlibC`.

lorentey: If we need to have `System` as a public import on Linux, then we should definitely import it there.

All I'm saying is that on Darwin, `CSystem` should only be imported `@_implementationOnly` (or not at all, given that we don't seem to need it).

Sources/System/Internals/Constants.swift
lorentey: Don't we need that dance in some files anyway, to call the functions or use types not exposed in CInterop? 

I imagine maintaining this would be more work than copy and pasting the right import incantation.

lorentey: Oh, unavailable names shadowing these is a great point! We can't easily resolve it by qualifying with the module name, because that's different between Darwin and Glibc. Bah.

If only we could say `import Darwin as _System` or something

Sources/System/Internals/Exports.swift
milseman: That will work until we do need it, at least. We wouldn't be able to surface `ResourceUsageInfo` as a raw-valued struct, but that's probably fine for the foreseeable future.

Sources/System/Internals/Constants.swift
milseman: We only do the import dance inside the `Internals` sub-folder (pending CTypes.swift in upcoming commit). I do agree that the dance alone is a toss up, but I really don't want to have to refer to explicit modules anywhere for constants.

PR #30, opened 2021-02-27
[DRAFT] Initial socket support

milseman:
Draft of `SocketDescriptor` and support for:

#### Prototyped

- `socket`, with domains (protocol/address families) and connection types (stream vs datagram)
- `send`/`recv`, with message flags
- `listen`
- `shutdown`
- Socket options (socket, ipv4, ipv6, and tcp) and `getsockopt` and `setsockopt`
- helpers, akin to `FileDescriptor`'s helpers

#### TODO

- `getaddrinfo`-like functionality
- `sockaddr_t` and friends, and syscalls taking `sockaddr_t`.
- unavailable-renamed entries for constants,
- Linux and Windows ports
- Better testing and mocking of socket options

#### Questions

- What is interruptible and what isn't?
- What should be `Codable`?
  - What is the policy, is it for same-process, cross-process, cross-system, or cross-platform serialization?
- Should we instead have a `TCP`, `IP`, `IPv6` namespace for socket options?
  - Then, `level` and `option_name` are in effect rolled into one parameter
  - Are the IPv4 options usable on IPv6 sockets? Should they have `v4` in their names?

#### Deferred

- network byte ordering operations (if any beyond stdlib)
- `gethostent`, `getnetent`, `getprotoent`, `getservent`, etc., unless needed for `sockaddr_t`
- failable check if a `FileDescriptor` really is a socket (`stat`)
- `Bool` or simple value-oriented API for options


Sources/System/Internals/Exports.swift
```diff
@@ -7,24 +7,25 @@
...
-@_implementationOnly import Darwin
+import Darwin
 #elseif os(Linux) || os(FreeBSD) || os(Android)
-@_implementationOnly import Glibc
+import CSystem
+import Glibc
 #elseif os(Windows)
-@_implementationOnly import ucrt
+import CSystem
+import ucrt
```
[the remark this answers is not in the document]
milseman: Hmm, is linking against WinSDK a problem? Could the shim module link against just `Ws2_32.lib`?

milseman:
Quick thought. We have

- SocketDescriptor
  - Option
    - OptionLevel
      - `tcp`
      - `ip`
      - `ipv6`
      - `socket`
    - `keepAlive` is `SO_KEEPALIVE` (perhaps we rename it to `socketKeepAlive`)
    - `tcpKeepAlive` is `TCP_KEEPALIVE`
    - ...

And `getOption`/`setOption` takes both a `Option.Level` and an `Option`, directly mirroring the C APIs (though these return values and are generic, so they're a little higher level).

An alternative would be something like
- SocketDescriptor
  - Option
    - Stores a level *and* and option-name
    - Level (as before, useful for `sendmsg`/`recvmsg` message headers)
    - Name (RR struct for `CInt`, formally `Option` above)
      - TCP: just a wrapper around `OptionName` to be a namespace
        -  `keepAlive` is `TCP_KEEPALIVE`
      - Socket: just a wrapper around `OptionName` to be a namespace
        - `keepAlive` is `SO_KEEPALIVE`
      - ...
    - `func tcp(_: OptionName.TCP) -> Option`
    - `func socket(_: OptionName.Socket) -> Option`
    - ...
    - `init(_ level: Option.Level, _ name: Option.Name)`, so people can always use raw values to make one

Then, `getOption/setOption` just take one parameter for both, and the call site looks like

```diff
- socket.getOption(.tcp, .tcpKeepAlive)
+ socket.getOption(.tcp(.keepAlive))
```

With the benefit of not requiring the caller to keep the level and specific option-name in sync, this API will do so automatically.

Sources/System/Internals/Exports.swift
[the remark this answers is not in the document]
compnerd: WinSDK is not a library but rather a hand rolled module that encompasses all of the Windows SDK.  Yes, the shim module can just link against `Ws2_32.lib` and that is sufficient.  I'm just worried about overlinking, but, I suppose that it should be easy to check.  The WinSDK module isn't great though and needs to be improved (it has been built up organically and tactically as we dont have good control over the SDK).

Sources/System/Sockets/SocketDescriptor.swift
```diff
@@ -0,0 +1,259 @@
...
+  /// The file descriptor for `self`.
+  @_alwaysEmitIntoClient
+  public var fileDescriptor: FileDescriptor {
+    FileDescriptor(rawValue: rawValue)
+  }
+
+  /// Treat `fd` as a socket descriptor, without checking with the operating
+  /// system that it actually refers to a socket
+  @_alwaysEmitIntoClient
+  public init(unchecked fd: FileDescriptor) {
```
kylemacomber, line 38: I wonder if `unchecked` is the right argument label to use here. It makes me think of `Range(uncheckedBounds:)`, in which case the checking is a bounds check precondition.

Sources/System/Sockets/SocketOperations.swift
```diff
@@ -0,0 +1,253 @@
...
+  /// Create an endpoint for communication.
+  ///
+  /// - Parameters:
+  ///    - domain: Select the protocol family which should be used for
+  ///      communication
+  ///    - type: Specify the semantics of communication
+  ///    - protocol: Specify a particular protocol to use with the socket.
+  ///      Normally, there is only one protocol for a particular connection
+  ///      type within a protocol family, so a default argument of `.default`
+  ///      is provided
```
kylemacomber, line 20: Formatting is off by one I think:

```suggestion
  ///   - domain: Select the protocol family which should be used for
  ///     communication
  ///   - type: Specify the semantics of communication
  ///   - protocol: Specify a particular protocol to use with the socket.
  ///     Normally, there is only one protocol for a particular connection
  ///     type within a protocol family, so a default argument of `.default`
  ///     is provided
```

Sources/System/Sockets/SocketOperations.swift
```diff
@@ -0,0 +1,253 @@
...
+
+  @usableFromInline
+  internal func _listen(backlog: Int) -> Result<(), Errno> {
+    nothingOrErrno(system_listen(self.rawValue, CInt(backlog)))
+  }
+
+  /// Send a message from a socket
+  ///
+  /// - Parameters:
+  ///   - buffear: The region of memory that contains the data being sent.
```
kylemacomber, line 85:
```suggestion
  ///   - buffer: The region of memory that contains the data being sent.
```

Sources/System/Internals/Exports.swift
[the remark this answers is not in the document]
milseman: Can we dynamically link to it or `dlsym` equivalent? I don't really know Windows best practices or conventions, but my recollection is that it was shockingly in favor of hot loading `.dll`s.

kylemacomber (commented):
Some superficial comments

Sources/System/Sockets/SocketDescriptor.swift
milseman: For `Range`, is that precondition later checked? Here, the failure mode is "recoverable", i.e. Errno is thrown, but it is still an operation that doesn't check that the result is valid until use. "unsafe" is flat out as this will be checked and does not produce UB.

Sources/Samples/ReverseResolve.swift
```diff
@@ -0,0 +1,77 @@
...
+      }
+      if numericService {
+        flags.insert(.numericService)
+      }
+      if datagram {
+        flags.insert(.datagram)
+      }
+      if scopeid {
+        flags.insert(.scopeIdentifier)
+      }
```
milseman, line 69: @natecook1000 , does the argument parser allow adding members to an OptionSet directly?

Sources/System/Sockets/SocketDescriptor.swift
```diff
@@ -0,0 +1,367 @@
...
+
+
+// TODO: @available(...)
+
+// TODO: Windows uses a SOCKET type; doesn't have file descriptor
+// equivalence
+
+/// TODO
+@frozen
+public struct SocketDescriptor: RawRepresentable, Hashable {
```
weissi, line 18: how are we later on doing kqueue/epoll where we can add sockets and pipes (and for kqueue also file descriptors (although they don't work like you'd expect to))?

Sources/Samples/ReverseResolve.swift
natecook1000: No - it would be really cool to have that as a feature, where a group of flags could all be stored together in an `OptionSet` type. You’d need to provide the metadata specifically, since those aren’t typically `CaseIterable` the way enums often are.

natecook1000: If this is something you’re using multiple places, you could move the flags to a separate type, vend the option set as a property, and then include it as an `@OptionGroup`.

Sources/System/Sockets/SocketDescriptor.swift
lorentey: No -- the `unchecked` in `Range(uncheckedBounds:)` indicates that failing to fulfill the unchecked precondition will lead to undefined behavior.

This is not the case here at all -- the difference between FileDescriptor and SocketDescriptor is a figment of our own imagination; mixing up the two will do no harm.

Sources/System/Sockets/SocketDescriptor.swift
lorentey: We can do that either by requiring the user to pass `socket.fileDescriptor` to these construct, or by adding direct overloads that take a SocketDescriptor, or by introducing a protocol for these things. Note that the latter option is highly unlikely, I think.

Sources/System/Internals/Exports.swift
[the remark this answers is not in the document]
lorentey: Yep, forcing unnecessary linkages on people would be a significant obstacle to adoption. (E.g., see corelibs-foundation.) There ought to be no reasonable objection against adopting swift-system.

I expect at some point we will have to take the step of splitting up this package into multiple system-dependent modules, with the top-level `SystemPackage` module reexporting each component.

Support for winsocks ought to be imported separately from the core system.

What modules (if any) need to be spun off is different on every platform, and the package should fully reflect that.

lorentey:
> What should be Codable?

I think we can make `SocketAddress.IPv4.Address`, `SocketAddress.IPv6.Address` and `SocketAddress.Port` codable. The addresses should probably serialize in String form.

Codable for `SocketAddress.IPv4`, `SocketAddress.IPv6` and `SocketAddress.Local` may also make sense, using a keyed container for the various components. However, I believe the underlying sockaddr_* structs may have system-dependent components that wouldn't necessarily survive serialization, so it's probably best not to do it.

None of the constant enums or other structs seem Codable to me.

Sources/System/Sockets/SocketDescriptor.swift
```diff
@@ -0,0 +1,367 @@
...
+        (.dataTruncated, ".dataTruncated"),
+        (.ancillaryTruncated, ".ancillaryTruncated"),
+      ]
+      return _buildDescription(descriptions)
+    }
+  }
+
+  /// Specify the part (or all) of a full-duplex connection to shutdown.
+  @frozen
+  public struct ShutdownKind: RawRepresentable, Hashable, Codable, CustomStringConvertible {
```
lorentey, line 333: I think we'll want to get rid of `Codable` here.

Sources/System/Sockets/SocketDescriptor.swift
weissi: @lorentey Right. The "issue" with this is that for `epoll` for example, you should only pass pipes/sockets/ttys etc. If you pass a regular file, then it won't work (and error, think even `EBADF` :) ). So given that we have some form of typing, it's kinda sad to remove it to go to epoll which internally does require the right type. But maybe this is a step too far for System

millenomi:
To answer your Codable question: Codable abstracts your serialization format from your type structure, and thus is suitable for all kinds of serialization, including cross-process and cross-machine. (One of its primary use cases is to parse JSON coming from arbitrary servers, for instance.)

Sources/System/Sockets/SocketDescriptor.swift
milseman: I haven't explored kqueue yet, but I suspect we'll follow a similar strategy here with a custom descriptor type. We might end up with rich type hierarchies for the different kinds of descriptors, so `epoll` would only receive pipe-family descriptors, but that's TBD.

If Swift had struct subtyping, this would all be way easier to model (as would many other things), but we can instead rely on conventions and protocols.

lorentey: The distinction between `SocketDescriptor` and `FileDescriptor` is just a namespacing convenience -- it's not real. If we add a `KqueueDescriptor`, it'll also be just a namespacing construct.

The primary goal of System is to eliminate the difficulty of calling syscalls from Swift, and especially to fix the associated memory safety issues.

Building abstractions that don't directly map to the underlying interfaces aggressively works against this goal -- however desirable they might be. Such abstractions are important (dare I say, crucial) to have, but they belong in higher layers such as NIO. 

Othe than ensuring memory safety, System cannot afford to be strongly opinionated about how people want to use syscalls; otherwise some folks will be forced to continue calling the unsafe C originals, and memory management / buffer overflow issues will be a problem forever.

Making it more difficult to call syscalls just to prevent EBADF/ENOTSOCK from getting thrown is definitely a non-goal to me. It's not unsafe to call epoll on a regular file; you'll just get a runtime error.

This is akin to calling `seek` on a terminal device or socket (in fact, that's even "worse", as `seek` may not even error out). 

It's not System's job to prevent people from doing stupid things with syscalls -- it's System's job to prevent people from continuing to make memory safety bugs.

Sources/System/Sockets/SocketOptions.swift
```diff
@@ -0,0 +1,605 @@
...
+    ///
+    /// The corresponding C constant is `SO_NREAD`.
+    @_alwaysEmitIntoClient
+    public static var getNumBytesToReceive: Option { Option(_SO_NREAD) }
+
+    /// Number of bytes written not yet sent by the protocol (get only).
+    ///
+    /// The corresponding C constant is `SO_NWRITE`.
+    @_alwaysEmitIntoClient
+    public static var getNumByteToSend: Option { Option(_SO_NWRITE) }
```
dfunckt, line 145:
```suggestion
    public static var getNumBytesToSend: Option { Option(_SO_NWRITE) }
```

Sources/System/Sockets/SocketOptions.swift
```diff
@@ -0,0 +1,605 @@
...
+
+    /// Linger on close if data present with timeout in seconds.
+    ///
+    /// The corresponding C constant is `SO_LINGER_SEC`.
+    @_alwaysEmitIntoClient
+    public static var longerSeconds: Option { Option(_SO_LINGER_SEC) }
```
dfunckt, line 151:
```suggestion
    public static var lingerSeconds: Option { Option(_SO_LINGER_SEC) }
```

dfunckt (commented):
Happened upon a couple typos and thought I'd mark them even though that part appears to be WIP.

milseman:
Rebased, fixed conflicts, and then extracted everything into a separate `SystemSockets` module. Still need to clean up some of the shared code, probably do a separate test target, and figure out how to get mocking hooked up.

karwa:
This would be great!

It's really awkward to use the system's socket API via the clang importer - for instance, it seems like every system uses a different name for the union property of an `in6_addr`; on Darwin it's `__u6_addr`, Glibc uses `__in6_u`, Windows does something different as well, but I can't figure out what...

FYI, `WebURL` includes [IPv4](https://karwa.github.io/swift-url/IPv4Address/) and [v6](https://karwa.github.io/swift-url/IPv6Address/) address types, with parsing and serialisation in Swift. The parsers support throwing errors, and are compatible with the obscure formats supported by `inet_aton`.

I'd be happy to contribute those to swift-system. IMO, it's nicer if you can avoid platform dependencies, because you can  guarantee more consistent behaviour, and there is not really anything about IP addresses which should be specific to any system. The implementation is fairly well isolated to a [single file](https://github.com/karwa/swift-url/blob/main/Sources/WebURL/IPAddress.swift), outside of a couple of ASCII constants and some pointer utilities for using tuples as fixed-size arrays.

Lukasa:
I’m not speaking for @milseman here, but while I think good IP address types are very valuable, they probably don’t belong in Swift System. In many ways, Swift System is not trying to paper over the differences between OSes, but is instead exposing them.

However, I definitely think some suitable currency types here would be valuable. They may belong in the standard library more than they belong here. Relatedly, currency types for the wider “sockaddr” notion would also be useful, as those types are drastically more complex than an IP address alone.

karwa:
Well, the API roadmap says:

> System aims to provide API in 3 areas:
> ...
> 2. Common types for library API at the systems level
> System hosts common types for systems level programming, enabling libraries built on top of System to use the same, well-> crafted types in their API.

So that's why I thought they might fit here; papering over OS differences is just a bonus. If I understand you correctly @Lukasa, you're saying that there may be room for a library below swift-system (maybe the stdlib), containing currency types defined by e.g. networking standards? That's an interesting idea.

Lukasa:
It does, but the README also says:

> ### Multi-platform not Cross-platform
>
> System is a multi-platform library, not a cross-platform one. It provides a separate set of APIs and behaviors on every supported platform, closely reflecting the underlying OS interfaces. A single import will pull in the native platform interfaces specific for the targeted OS.
> 
> Our immediate goal is to simplify building cross-platform libraries and applications such as SwiftNIO and SwiftPM. System does not eliminate the need for `#if os()` conditionals to implement cross-platform abstractions, but it does make it safer and more expressive to fill out the platform-specific parts.

Naturally these two ideas are in tension, but I think it would be good to holistically consider whether swift-system is the right place to define an IPAddress type, or whether it's simply somewhere that should _use_ it.

milseman:
I do want to resuscitate this and figure out how to move forwards.

@karwa thanks for the offer for IP address parsing code. Is this a desire to avoid making a syscall, or are there concerns about platform availability? Are you able to support more API with a native implementation? We have construction of address from string via `pton` et al.

System is surfacing platform differences, but in a way that should make it easier to write platform-conscious code (including working in portable subsets). Sort of like the `P` is POSIX, it should be possible to do (even if not ideally) without crazy dangerous unsafe pointers. At the same time, some concepts like `FilePath` are so important and interwoven with libraries that do wish to work with portable interfaces that we invest significant amounts of library code and API design into supporting.

Lukasa:
> If we remove RawRepresentable then rawValue could be replaced with description but I dont find a better candidate for init?(rawValue: RawValue). @Lukasa What would your suggestion be to replace that in a way that plays nicely with the Swift stdlib and is idiomatic (and doesn't reinvent the wheel)?

`init(string:)` is fine.

Concretely, my objection here is that `"127.0.0.1"` _isn't_ the raw representation of the IP address that it represents: `UInt32(0x7F000001)` is. `"127.0.0.1"` is one of many possible string representations of the same IP address. Users won't believe we're just passing strings around, because IP addresses aren't strings: they're numbers. This is the reason not to move Strings around: it has nothing to do with ARC, and everything to do with representing what an IP address actually is.

> I am aware of the ABI aspect and I think we should freeze the enums we ship with this library.

That's good, because we don't have a choice: Swift packages literally cannot have open enums in them.

> Again I point to POSIXError in Foundation for Linux, Windows and Darwin (which I contributed to) as examples for using enums for C constants.

`POSIXError` is a `struct`: https://github.com/apple/swift-corelibs-foundation/blob/599c05d83183454bea653b8843a9e26ca84f4a4c/Sources/Foundation/NSError.swift#L980.

Additionally, socket options in your design aren't constants: they have associated values, because they encode the type of the value. This is a good part of your design, and I want to see it persist: dealing with non `CInt`-sized socket options is painful. I'm just saying that enums is not a good fit for this design: structs are a perfect fit.

We don't have to "litter the namespace" with them: you can define them in an uninhabited `enum` and use that to namespace them. They just shouldn't be cases.

> Yes, but we want to make it as verbose as all the other `unsafe` APIs in Swift (e.g. `withUnsafePointer()`), not a simple `.init(rawValue: sockaddr_in6)`.

Why? It's not unsafe.

Lukasa:
I don't think I did.

Here's the code directly from your module (as of [eba8454e](https://github.com/PureSwift/swift-system/blob/eba8454e56c16e6d0caad2999035941c4958bc39/Sources/System/SocketOption.swift#L11-L48)):

```swift
public protocol SocketOption {
    
    associatedtype ID: SocketOptionID
    
    var id: ID { get }
    
    func withUnsafeBytes<T>(_: ((UnsafeRawBufferPointer) -> (T))) -> T
}

public enum GenericSocketOption: SocketOption, Equatable, Hashable {
    
    public typealias ID = GenericSocketOptionID
    
    case debug(Bool)
    case keepAlive(Bool)
    
    @_alwaysEmitIntoClient
    public var id: ID {
        switch self {
        case .debug: return .debug
        case .keepAlive: return .keepAlive
        }
    }
    
    @_alwaysEmitIntoClient
    public func withUnsafeBytes<T>(_ pointer: ((UnsafeRawBufferPointer) -> (T))) -> T {
        switch self {
        case let .debug(value):
            return Swift.withUnsafeBytes(of: value.cInt) { bufferPointer in
                pointer(bufferPointer)
            }
        case let .keepAlive(value):
            return Swift.withUnsafeBytes(of: value.cInt) { bufferPointer in
                pointer(bufferPointer)
            }
        }
    }
}
```

My point here is that `GenericSocketOption` is not something that this library can _ever_ evolve. We will get one shot to define the type, and then it'll be frozen in time forever. This requires us to audit for essentially all possible socket options associated with SOL_SOCKET on supported platforms and define as many of them as possible, or we'll lose the ability to refer to them by a shorthand. If we ever _did_ want to add something to this library that added more options for SOL_SOCKET, we'd have to define a brand new type for them.

As to POSIXErrorCode, enums in the standard library _can_ add new cases over time, so my objection does not apply to them. POSIXErrorCode, notably, is _not_ a frozen enum.

karwa:
> Thats fair, I guess its a personal preference but I do see the logic of what you are saying. My logic was that since its the preferred way to construct the type (with the C type also able to instantiate it, but via more verbose unsafe initializer), then for the end user the fact we are not moving a String around (to avoid ARC) and converting it at the last minute to C types is merely an optimization and implementation detail. Most usage of this API in Swift will be creating the address from String and most of the usage of the C struct will predictably be used with interfacing with C stdlib, which this module is trying to wrap and discourage direct usage

You don't have to construct an IP address from a String. With the WebURL IPv4Address type you can also write:

```swift
let addr = IPv4Address(octets: (127, 0, 0, 1))
```

This has some uses for addresses you know at compile-time (loopback addresses, perhaps special services). Since it doesn't require parsing, it isn't a failable initialiser. IPv6 addresses offer the same thing, but they're easier to get wrong because they're bigger.

PR #32, opened 2021-03-14
Emit withPlatformString as an availability workaround for open

milseman:
Unfortunately introduces a warning until the compiler can give us a way to call deprecated methods.

PR #33, opened 2021-03-15
[Sketch] `sysconf`, `pathconf`, `fpathconf`

milseman:
Just a sketchup of support for POSIX system variable configuration. Does not include the more complete, and platform-specific, `sysctl`.

NOTE: This is based on top of the cleanup and fixes in https://github.com/apple/swift-system/pull/32.


milseman:
@lorentey so I'm interested in landing these sketches. The API design question here is how best to surface global calls. Here we have a `enum SystemConfig {}` and a static `get(_ name: Name) -> Int` method for `sysconf`. An alternative would be some kind of shared namespace for all of these kinds of global calls, not just for configuration.

lorentey (approved):
I like this, but I have many comments! It would be great if we could unify all these under a single API umbrella, but I have some doubts if that's the right direction.

PR #37, opened 2021-03-16
Internals: add Windows error code equivalences

compnerd:
Windows does not have the same set of error codes as a POSIX system.
However, the WinSock subsystem provides a set of error codes which are
meant to provide a means for handling similar error scenarios with
sockets.  This additionally adds an additional error code which allows
us to add back a previously ignored error code.

### DIFF
--- a/Sources/System/Errno.swift
+++ b/Sources/System/Errno.swift
@@ -634,7 +634,6 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @available(*, unavailable, renamed: "socketTypeNotSupported")
   public static var ESOCKTNOSUPPORT: Errno { socketTypeNotSupported }
 
-#if !os(Windows)
   /// Not supported.
   ///
   /// The attempted operation isn't supported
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
Note that this is not entirely semantically equivalent.  The errors mapped here are not valid in all contexts but rather are limited to the context of calls to the WinSock subsystem.

compnerd:
@swift-ci please test

milseman (approved):
LGTM. Out of curiosity, how was `ESOCKTNOSUPPORT` compiling before this change?

compnerd:
I don't believe that it was.  I believe that it was added after the initial port, and because there is no CI, we never noticed that we broke the build.  There were a few instances of that.  However, once a CMake based build is in place, I can easily setup a nightly build job to build Swift System on Windows, and should even be able to kick off one-offs with a patch if desired.

milseman:
Ah, that was because we were hand-declaring them with the int literals instead of the module constants: https://github.com/apple/swift-system/commit/4ffcc62e9b6bb3bee3dcf1d1dc03c5684cd26e6a#diff-0ee5719a7e727f19909c8ae438eb8ad3e0cdaa454fc4e1e51cea56f34feec8c1R214

milseman:
@swift-ci please test linux platform

PR #38, opened 2021-03-16
build: add a CMake based build system

compnerd:
This will allow using Swift System in the toolchain stack, allowing us
to bootstrap a toolchain.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,29 @@
+#[[
+This source file is part of the Swift System open source project
+
+Copyright (c) 2020 Apple Inc. and the Swift System project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+cmake_minimum_required(VERSION 3.16.0)
+project(swift-system
+  LANGUAGES C Swift)
+
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+
+include(SwiftSupport)
+
+add_subdirectory(Sources)
+
+get_property(SWIFT_SYSTEM_EXPORTS GLOBAL PROPERTY SWIFT_SYSTEM_EXPORTS)
+export(TARGETS ${SWIFT_SYSTEM_EXPORTS}
+  NAMESPACE SwiftSystem::
+  FILE swift-system-config.cmake
+  EXPORT_LINK_INTERFACE_LIBRARIES)
--- a/Sources/System/CMakeLists.txt
+++ b/Sources/System/CMakeLists.txt
@@ -1,0 +1,40 @@
+#[[
+This source file is part of the Swift System open source project
+
+Copyright (c) 2020 Apple Inc. and the Swift System project authors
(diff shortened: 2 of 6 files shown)


Sources/System/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  Errno.swift
+  FileDescriptor.swift
+  FileHelpers.swift
+  FileOperations.swift
+  FilePermissions.swift
+  PlatformString.swift
+  SystemString.swift
+  Util.swift
+  UtilConsumers.swift)
+target_sources(System PRIVATE
```
milseman, line 20: What does `PRIVATE` mean?

Sources/System/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+This source file is part of the Swift System open source project
+
+Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+add_library(System
+  Errno.swift
```
milseman, line 11: I don't suppose there's any reasonable way to say `*.swift` within a  directory? SPM already dictates that our source code matches our targets

CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+#[[
+This source file is part of the Swift System open source project
+
+Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
```
milseman, line 8: Change comment to System's header comment.

CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+include(SwiftSupport)
+
+add_subdirectory(Sources)
+
+get_property(SWIFT_SYSTEM_EXPORTS GLOBAL PROPERTY SWIFT_SYSTEM_EXPORTS)
+export(TARGETS ${SWIFT_SYSTEM_EXPORTS}
+  NAMESPACE SwiftSystem::
+  FILE swift-system-config.cmake
+  EXPORT_LINK_INTERFACE_LIBRARIES)
```
milseman, line 29: What does this `export` do?

Sources/System/CMakeLists.txt
compnerd: In the case of `target_sources` the `PRIVATE` indicates that the listed sources are not meant to be used to populate `INTERFACE_SOURCES` for the target.  This means that the sources listed here do not get propagated to users of this library.

Sources/System/CMakeLists.txt
compnerd: No, that is the way that CMake is designed - to be explicit.  If you try to use a glob, any changes to the source tree structure will not be reflected in the build until a clean build is done.

CMakeLists.txt
compnerd: Sure, that should be easy to do.

CMakeLists.txt
compnerd: This export is the trick that enables wiring up the entire toolchain build together.  It creates a `swift-system-config.cmake` which can then be used for any subsequent build to just grab the build artifacts from the build tree without having to install and create a single distribution image and incrementally add content into it.  Once everything is built and verified to work together, you just create a single image that you intend to distribute.

Sources/System/CMakeLists.txt
milseman: But do top-level sources? Unlike the `Internals` folder, this is just a way of organizing `FilePath` code together for navigation purposes. There's still API inside.

milseman:
@swift-ci please test

Sources/System/CMakeLists.txt
compnerd: ah, I see your point.  No, the top-level sources are treated as private by default.  I only did the split `target_sources` for reading.  We can even merge them into a single call if you prefer.  That is a stylistic thing.

milseman: I have no opinions here, just hoping to understand CMake a little bit better :-)

compnerd:
@swift-ci please test

PR #40, opened 2021-03-20
Internals: replace `CModeT` with `CInterop.Mode`

compnerd:
The typealias was renamed but the reference to it was not fully removed.
This updates the instance enabling building System on Windows.


compnerd:
@swift-ci please test

PR #41, opened 2021-03-20
FilePath: enable building on Windows

compnerd:
When building on Windows, we rely on the `withPlatformString` to convert
the strings to the native platform encoding (UTF-16).  The public
`FilePath.withPlatformString` is marked as `@_alwaysEmitIntoClient` and
references the `_withPlatformString` which is marked as `internal`.
Annotate the function as `@usableFromInline` to enable the inlined
method to reference this function.

### DIFF
--- a/Sources/System/FilePath/FilePathComponents.swift
+++ b/Sources/System/FilePath/FilePathComponents.swift
@@ -190,6 +190,7 @@ extension FilePath.Root: _PathSlice {
   }
 }
 extension FilePath: _PlatformStringable {
+  @usableFromInline
   func _withPlatformString<Result>(_ body: (UnsafePointer<CInterop.PlatformChar>) throws -> Result) rethrows -> Result {
     try _storage.withPlatformString(body)
   }


compnerd:
@swift-ci please test

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -190,6 +190,7 @@ extension FilePath.Root: _PathSlice {
   }
 }
 extension FilePath: _PlatformStringable {
+  @usableFromInline
```
milseman, line 193: @lorentey we will need to version this for `Darwin` binary distribution, but otherwise I think this is an ok workaround.

milseman (approved):

Sources/System/FilePath/FilePathComponents.swift
lorentey: Adding `@usableFromInline` would introduce this as ABI, which we don't actually need/want to do on Darwin. (Its availability wouldn't match the rest of FilePath, making it inconvenient.)

I'll try to figure out an alternative solution. I'm not sure why the caller needs to be marked `@_alwaysEmitIntoClient` on Windows, for example.

PR #43, opened 2021-03-22
[Sketch] openat, fsync, and sync

milseman:


milseman:
This is close to being mergeable, but still has the global system call design question as https://github.com/apple/swift-system/pull/33

Sources/System/FileOperations.swift
```diff
@@ -371,3 +371,109 @@ extension FileDescriptor {
     fatalError("Not implemented")
   }
 }
+
+extension FileDescriptor {
+  /// Opens or creates a file for reading or writing.
+  ///
+  /// - Parameters:
+  ///   - path: The location of the file to open.
+  ///   - at: if `path` is relative, treat it as relative to this file descriptor
```
lorentey, line 380: Nit: we should not use prepositions as variable names.

```suggestion
  ///   - base: if `path` is relative, treat it as relative to this file descriptor
```

Or we could just call it `fd`.

Sources/System/FileOperations.swift
```diff
@@ -371,3 +371,109 @@ extension FileDescriptor {
...
+  ///     if it throws ``Errno/interrupted``.
+  ///     The default is `true`.
+  ///     Pass `false` to try only once and throw an error upon interruption.
+  /// - Returns: A file descriptor for the open file
+  ///
+  /// The corresponding C function is `openat`.
+  @_alwaysEmitIntoClient
+  public static func open(
+    _ path: FilePath,
+    at: FileDescriptor,
```
lorentey, line 395:
```suggestion
    at base: FileDescriptor,
```

(etc, throughout this PR.)

lorentey (approved):
💯 🚢

rauhul:
@milseman @lorentey Do we want this to be called `open(_:at:...)` or `open(_:relativeTo:...)`. Whatever precedent is set here, should be followed by my `fstatat` functions

lorentey:
I'd be happy with either option! I like `relativeTo:`, too.

milseman:
I much prefer `relativeTo:` personally as well, as I otherwise would do a double-take in the man page or docs to make sure `at` means what it means in `openat(2)`.

PR #44, opened 2021-04-06
[SKETCH] Add in fcntl raw, and higher-level support.

milseman:
Add in support for fcntl and all the raw commands and flags it supports. Add in Swiftier representations of common and important invocations. 

### System Sketch

New design patterns:
  - We provide rawer, but complete, `FileDescriptor.fcntl()` and `FileDescriptor.Command` interfaces and wrappers.
  - We provide Swiftier type-safe interfaces for a common and important subset.
  - `FileDescriptor.FileLock` is our first RR wrapper of a real C struct.

Externals:
  - PID, and PIDOrGPID.
  - `flock` to coordinate design with `get/setLock()`.
  - `fsync` to coordinate design with `F_FULLFSYNC` etc.

Unknowns:
  - Headers have even more locking commands/variants, should we promote these to the Swifty APIs?
  - Is `Errno.resourceTemporarilyUnavailable` too removed from `EAGAIN`? Should we introduce `Errno.again` alias?
  - Are there other Swiftier interfaces we should provide?

Testing needs:
  - Normal mocking tests
  - Sample code, especially for provided Swiftier interfaces

### DIFF
--- a/Sources/System/FileControl.swift
+++ b/Sources/System/FileControl.swift
@@ -1,0 +1,216 @@
+/*
+ This source file is part of the Swift System open source project
+
+ Copyright (c) 2021 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+// Strongly typed, Swifty interfaces to the most common and useful `fcntl`
+// commands.
+
+extension FileDescriptor {
+  /// Get the flags associated with this file descriptor
+  ///
+  /// The corresponding C function is `fcntl` with the `F_GETFD` command.
+  @_alwaysEmitIntoClient
+  public func getFlags() throws -> Flags {
+    try Flags(rawValue: fcntl(.getFlags))
+  }
+
+  /// Set the file descriptor flags.
+  ///
+  /// The corresponding C function is `fcntl` with the `F_SETFD` command.
+  @_alwaysEmitIntoClient
+  public func setFlags(_ value: Flags) throws {
+    _ = try fcntl(.setFlags, value.rawValue)
+  }
+
+  /// Get descriptor status flags.
+  ///
+  /// The corresponding C function is `fcntl` with the `F_GETFL` command.
+  @_alwaysEmitIntoClient
+  public func getStatusFlags() throws -> StatusFlags {
+    try StatusFlags(rawValue: fcntl(.getStatusFlags))
+  }
+
+  /// Set descriptor status flags.
+  ///
+  /// The corresponding C function is `fcntl` with the `F_SETFL` command.
+  @_alwaysEmitIntoClient
+  public func setStatusFlags(_ flags: StatusFlags) throws {
+    _ = try fcntl(.setStatusFlags, flags.rawValue)
+  }
+}
+
+// FIXME: We want a better approach for this one.
+extension FileDescriptor {
+  /// TODO: Flesh out PID work and see if there's a better, common formulation
+  /// of the process-or-group id concept.
+  @frozen
+  public struct PIDOrPGID: RawRepresentable {
+    @_alwaysEmitIntoClient
+    public let rawValue: CInt
+
+    @_alwaysEmitIntoClient
+    public init(rawValue: CInt) { self.rawValue = rawValue }
+
+    /// TODO: PID type
+    @_alwaysEmitIntoClient
+    public var asPID: CInt? { rawValue >= 0 ? rawValue : nil }
+
+    /// TODO: PGID type
+    @_alwaysEmitIntoClient
+    public var asPositiveGroupID: CInt? {
+      rawValue >= 0 ? nil : -rawValue
+    }
+
+    /// TODO: PID type
+    @_alwaysEmitIntoClient
+    public init(pid id: CInt) {
+      precondition(id >= 0)
+      self.init(rawValue: id)
+    }
+
+    /// TODO: PGID type
+    @_alwaysEmitIntoClient
+    public init(positiveGroupID id: CInt) {
+      precondition(id >= 0)
+      self.init(rawValue: -id)
+    }
+  }
+
+  /// Get the process ID or process group currently receiv-
+  /// ing SIGIO and SIGURG signals.
+  ///
+  /// The corresponding C function is `fcntl` with the `F_GETOWN` command.
+  @_alwaysEmitIntoClient
+  public func getOwner() throws -> PIDOrPGID {
+    try PIDOrPGID(rawValue: fcntl(.getOwner))
+  }
+
+  /// Set the process or process group to receive SIGIO and
+  /// SIGURG signals.
+  ///
+  /// The corresponding C function is `fcntl` with the `F_SETOWN` command.
+  @_alwaysEmitIntoClient
+  public func setOwner(_ id: PIDOrPGID) throws {
+    _ = try fcntl(.setOwner, id.rawValue)
+  }
+}
+
+extension FileDescriptor {
+  /// Duplicate this file descriptor and return the newly created copy.
+  ///
+  /// - Parameters:
+  ///   - `minRawValue`: A lower bound on the new file descriptor's raw value.
+  ///   - `closeOnExec`: Whether the new descriptor's `closeOnExec` flag is set.
+  /// - Returns: The lowest numbered available descriptor whose raw value is
+  ///     greater than or equal to `minRawValue`.
+  ///
+  /// File descriptors are merely references to some underlying system resource.
+  /// The system does not distinguish between the original and the new file
+  /// descriptor in any way. For example, read, write and seek operations on
+  /// one of them also affect the logical file position in the other, and
+  /// append mode, non-blocking I/O and asynchronous I/O options are shared
+  /// between the references. If a separate pointer into the file is desired,
+  /// a different object reference to the file must be obtained by issuing an
+  /// additional call to `open`.
+  ///
+  /// However, each file descriptor maintains its own close-on-exec flag.
+  ///
+  /// The corresponding C functions are `fcntl` with `F_DUPFD` and
+  /// `F_DUPFD_CLOEXEC`.
+  @_alwaysEmitIntoClient
+  public func duplicate(
+    minRawValue: CInt, closeOnExec: Bool
+  ) throws -> FileDescriptor {
+    let cmd: Command = closeOnExec ? .duplicateCloseOnExec : .duplicate
+    return try FileDescriptor(rawValue: fcntl(cmd, minRawValue))
+  }
+
+  // TODO: What is the firmlink thing about?
+
+  /// Get the path of the file descriptor
+  ///
+  /// - Parameters:
+  ///   - `noFirmLink`: Get the non firmlinked path of the file descriptor.
+  ///
+  /// The corresponding C functions are `fcntl` with `F_GETPATH` and
+  /// `F_GETPATH_NOFIRMLINK`.
+  public func getPath(noFirmLink: Bool = false) throws -> FilePath {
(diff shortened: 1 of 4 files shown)


milseman:
In light of the.... poorly imagined aspects of POSIX `fcntl` locks, we might drop the Swiftier lock interface, pending on how `flock` turns out.

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+    ///
+    /// The corresponding C constant is `F_GETPATH`.
+    @_alwaysEmitIntoClient
+    public static var getPath: Command { Command(F_GETPATH) }
+
+    /// Fsync + ask the drive to flush to the media.
+    ///
+    /// The corresponding C constant is `F_FULLFSYNC`.
+    @_alwaysEmitIntoClient
+    public static var fullFsync: Command { Command(F_FULLFSYNC) }
```
lorentey, line 386: What are we going to call `fsync(2)`? I think I'd suggest `FileDescriptor.sync` or `synchronize`. We should use the same name here.

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+    @_alwaysEmitIntoClient
+    public static var pathPackageCheck: Command {
+      Command(F_PATHPKG_CHECK)
+    }
+
+    /// "Freeze" all fs operations.
+    ///
+    /// The corresponding C constant is `F_FREEZE_FS`.
+    @_alwaysEmitIntoClient
+    public static var freezeFS: Command { Command(F_FREEZE_FS) }
```
lorentey, line 400: Have we used FS to mean filesystem before?

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+    @_alwaysEmitIntoClient
+    public static var getSignatureInfo: Command {
+      Command(F_GETSIGSINFO)
+    }
+
+    /// Shared or read lock.
+    ///
+    /// The corresponding C constant is `F_RDLCK`.
+    @_alwaysEmitIntoClient
+    public static var readLock: Command { Command(F_RDLCK) }
```
lorentey, line 648: Wait, are `F_RDLCK`/`F_WRLCK`/`F_UNLCK` really fcntl commands?

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+
+    /// Exclusive or write lock.
+    ///
+    /// The corresponding C constant is `F_WRLCK`.
+    @_alwaysEmitIntoClient
+    public static var writeLock: Command { Command(F_WRLCK) }
+
+    /// Allocate contigious space.
+    ///
+    /// TODO: This is actually a flag for the PREALLOCATE struct...
```
lorentey, line 664: Ah, I think these will need a separate enum-like struct. :-/

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+    /// The corresponding C constant is `F_ALLOCATEALL`.
+    @_alwaysEmitIntoClient
+    public static var allocateAll: Command {
+      Command(F_ALLOCATEALL)
+    }
+
+    /// Make it past all of the SEEK pos modes so that.
+    ///
+    /// TODO: The above is a direct quote from the header, figure out what it
+    /// means
```
lorentey, line 685: That's brilliant. The man page has some usable docs:

> Allocate from the physical end of file.  In this case, `fst_length` indicates the number of newly allocated bytes desired.

lorentey (dismissed):

Sources/System/FileControl.swift
```diff
@@ -0,0 +1,216 @@
...
+    _ = try fcntl(.setStatusFlags, flags.rawValue)
+  }
+}
+
+// FIXME: We want a better approach for this one.
+extension FileDescriptor {
+  /// TODO: Flesh out PID work and see if there's a better, common formulation
+  /// of the process-or-group id concept.
+  @frozen
+  public struct PIDOrPGID: RawRepresentable {
```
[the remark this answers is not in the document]
milseman: Right, this is a cross-cutting concern between this sketch and how we end up handling processes, hence the TODO. I called this out as an "External", meaning that will have to be figured out (or punted) before this can be merged.

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+  public struct FileLock: RawRepresentable {
+    @_alwaysEmitIntoClient
+    public var rawValue: CInterop.FileLock
+
+    @_alwaysEmitIntoClient
+    public init(rawValue: CInterop.FileLock) { self.rawValue = rawValue }
+
+    /// The type of the lock.
+    @frozen
+    public struct LockType: RawRepresentable, Hashable {
```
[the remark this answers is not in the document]
milseman: I really try to avoid having a `.Type` that isn't the metatype of a generic type parameter, but I don't know how confusing this would end up being in practice.

Sources/System/FileControl.swift
```diff
@@ -0,0 +1,216 @@
...
+  /// region, the it will be interrupted if the signal handler has not
+  /// specified the SA_RESTART (see sigaction(2)).
+  ///
+  /// The corresponding C function is `fcntl` with `F_SETLK`.
+  @_alwaysEmitIntoClient
+  public func setLock(
+    to lock: FileLock,
+    waitIfBlocked: Bool = false,
+    retryOnInterrupt: Bool = true
+  ) throws {
```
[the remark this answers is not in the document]
milseman: Both those sounds like fine suggestions. Honestly, I think we'd go with only having the `flock()` equivalent directly on `FileDescriptor`. We can provide this functionality through the general `fcntl` fallback and not privilege POSIX-style record locks. From the man page:

> This interface follows the completely stupid semantics of System V and IEEE Std 1003.1-1988 (``POSIX.1'') that require that all locks  associated with a file for a given process are removed when any file descriptor for that file is closed by that process.  This semantic  means that applications must be aware of any files that a subroutine library may access.  For example if an application for updating the  password file locks the password file database while making the update, and then calls getpwname(3) to retrieve a record, the lock will  be lost because getpwname(3) opens, reads, and closes the password database.  The database close will release all locks that the process  has associated with the database, even if the library routine never requested a lock on the database.  Another minor semantic problem  with this interface is that locks are not inherited by a child process created using the fork(2) function.  The flock(2) interface has  much more rational last close semantics and allows locks to be inherited by child processes.  Flock(2) is recommended for applications  that want to ensure the integrity of their locks when using library routines or wish to pass locks to their children.  Note that  flock(2) and fcntl(2) locks may be safely used concurrently.

Sources/System/FileControlRaw.swift
milseman: We haven't really expressed the concept of file system in names at all, IIRC.

Sources/System/FileControlRaw.swift
milseman: They're in the "Command" namespace but they're used as parameters. These grab-bag APIs don't have clear orthogonal namespaces. We could consider separating them out. There's a tradeoff between spending our design/weirdness budget on making these APIs better and directly surfacing the interfaces as-is. For now, I've been promoting the most important/common to `FileDescirptor` namespace, while the obscure or specialized options try to mirror `fcntl` as closely as possible.

Sources/System/FileControl.swift
```diff
@@ -0,0 +1,216 @@
...
+  // TODO: What is the firmlink thing about?
+
+  /// Get the path of the file descriptor
+  ///
+  /// - Parameters:
+  ///   - `noFirmLink`: Get the non firmlinked path of the file descriptor.
+  ///
+  /// The corresponding C functions are `fcntl` with `F_GETPATH` and
+  /// `F_GETPATH_NOFIRMLINK`.
+  public func getPath(noFirmLink: Bool = false) throws -> FilePath {
```
weissi, line 142: this will need `#if os(Darwin...`. On Linux you do this by running `readlink(/proc/self/FD_NUMBER)`

Sources/System/FileControl.swift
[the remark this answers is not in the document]
weissi: POSIX-style record locks are crazy! Don't think we should offer them :). They are identified _per pid & VFS node_. Ie. if two threads in one process open the same file and lock distinct parts that'll succeed. So correctly, both hold a lock to distinct parts of the same file. But if one of the threads then closes its fd (which is different from the other thread's fd), then _both_ locks go away. Ie the other threads _thinks_ it still has it locked but it doesn't.

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+    public static var nonBlocking: StatusFlags { StatusFlags(O_NONBLOCK) }
+
+    /// Force each write to append at the end of file; corre-
+    /// sponds to `OpenOptions.append`.
+    ///
+    /// The corresponding C constant is `O_APPEND`.
+    @_alwaysEmitIntoClient
+    public static var append: StatusFlags { StatusFlags(O_APPEND) }
+
+    /// Enable the SIGIO signal to be sent to the process
```
weissi, line 69: does anybody still use this? Do we really want to offer it?

Sources/System/FileControlRaw.swift
```diff
@@ -0,0 +1,726 @@
...
+
+    /// No SIGPIPE generated on EPIPE.
+    ///
+    /// The corresponding C constant is `F_SETNOSIGPIPE`.
+    @_alwaysEmitIntoClient
+    public static var setNoSigPipe: Command {
+      Command(F_SETNOSIGPIPE)
+    }
+
+    /// Status of SIGPIPE for this fd.
```
weissi, line 511: Linux doesn't have that either IIRC

Sources/System/FileControl.swift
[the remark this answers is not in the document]
milseman: I'm going to punt this type and API from the fcntl work, to be resolved when we handle processes, `kill`, etc.

Sources/System/FileControl.swift
```diff
@@ -0,0 +1,216 @@
...
+  /// `F_DUPFD_CLOEXEC`.
+  @_alwaysEmitIntoClient
+  public func duplicate(
+    minRawValue: CInt, closeOnExec: Bool
+  ) throws -> FileDescriptor {
+    let cmd: Command = closeOnExec ? .duplicateCloseOnExec : .duplicate
+    return try FileDescriptor(rawValue: fcntl(cmd, minRawValue))
+  }
+
+  // TODO: What is the firmlink thing about?
```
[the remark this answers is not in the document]
milseman: What do you think the default should be? Vanilla or no-firmlink?

Sources/System/FileControlRaw.swift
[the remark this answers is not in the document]
milseman: Per man page guidance, I'm removing this in favor of `flock(2)` support. There may still be the raw commands / constants if someone really wants it, but no need for these types.

milseman:
Rebased on top of https://github.com/apple/swift-system/pull/111 and closer to being merge-worthy.

Needs some simple tests, some differences between Linux and Darwin dissected, and version annotations.

milseman:
@swift-ci please test

milseman:
@swift-ci please test

Sources/System/FileControlRaw.swift
milseman: Went with `FileSystem`

milseman:
@swift-ci please test

milseman:
Early draft of API doc: https://gist.github.com/milseman/2e59c10299f04c682c4090c4d2ecc0e0

PR #48, opened 2021-05-11
FilePath.Component.Kind is meant to be frozen

milseman:


lorentey (approved):
Looks good to me. Forcing people to cover `@unknown` cases seems like a bad idea in this case.

In the unlikely case a new component kind is ever introduced, we can simply add a new enum & property to cover it.

I presume we don't want this to cover device references like `CON` or `LPT4.txt` on Windows, right?

milseman:
Legacy DOS devices are definitely not for this area of the library. Kind doesn't differentiate between folders and files, for example.

weissi:
@milseman / @lorentey We should fix the documentation then, right? Currently, `.regular` is

>    /// A file or directory name
>    case regular

and either untrue of misleading for  `/dev/tty`, UNIX domain sockets and much more.

Also CC @compnerd , any particular issues on Windows that we'd need to be aware of, the likes of `C:\CON\CON` or so is what I'm thinking.

lorentey:
This is not `stat`; `/dev/tty` is a file- or directory name.

weissi:
> This is not `stat`; `/dev/tty` is a file- or directory name.

So why "file- or directory name" or not just "file name" then if we buy into that? As in, what makes the directory special enough the mention it as separate from a file that doesn't apply to say a block device node?

lorentey:
I expect the answer is that directories can sometimes be distinguished syntactically. For example, in `/dev/tty`, it is pretty clear that `dev` has the role of a directory.

For me personally, the word "filename" covers directory names, too, but I don't have an issue with consistently clarifying it.

PR #51, opened 2021-06-08
Fix building on Windows

stevapple:
Nothing but as shown in the title.

The base for apple/swift-tools-support-core#219, apple/swift-tools-support-core#221

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,11 +12,19 @@
 
 import PackageDescription
 
-let targets: [PackageDescription.Target] = [
+var windowsPlatform: [Platform] = []
+#if os(Windows)
+windowsPlatform.append(.windows)
+#endif
+
+let targets: [Target] = [
   .target(
     name: "SystemPackage",
     dependencies: ["CSystem"],
     path: "Sources/System",
+    cSettings: [
+      .define("_CRT_SECURE_NO_WARNINGS", .when(platforms: windowsPlatform)),
+    ],
     swiftSettings: [
       .define("SYSTEM_PACKAGE"),
       .define("ENABLE_MOCKING", .when(configuration: .debug))
--- a/Package.swift
+++ b/Package.swift
@@ -12,39 +12,32 @@
 
 import PackageDescription
 
-var windowsPlatform: [Platform] = []
-#if os(Windows)
-windowsPlatform.append(.windows)
-#endif
-
-let targets: [Target] = [
-  .target(
-    name: "SystemPackage",
-    dependencies: ["CSystem"],
-    path: "Sources/System",
-    cSettings: [
-      .define("_CRT_SECURE_NO_WARNINGS", .when(platforms: windowsPlatform)),
-    ],
-    swiftSettings: [
-      .define("SYSTEM_PACKAGE"),
-      .define("ENABLE_MOCKING", .when(configuration: .debug))
-    ]),
-  .target(
-    name: "CSystem",
-    dependencies: []),
-  .testTarget(
-    name: "SystemTests",
-    dependencies: ["SystemPackage"],
-    swiftSettings: [
-      .define("SYSTEM_PACKAGE")
-    ]),
-]
-
 let package = Package(
     name: "swift-system",
     products: [
         .library(name: "SystemPackage", targets: ["SystemPackage"]),
     ],
     dependencies: [],
-    targets: targets
+    targets: [
+      .target(
+        name: "CSystem",
+        dependencies: []),
+      .target(
+        name: "SystemPackage",
+        dependencies: ["CSystem"],
+        path: "Sources/System",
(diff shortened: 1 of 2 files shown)


milseman:
@compnerd

Package.swift
```diff
@@ -12,11 +12,19 @@
 
 import PackageDescription
 
-let targets: [PackageDescription.Target] = [
+var windowsPlatform: [Platform] = []
+#if os(Windows)
+windowsPlatform.append(.windows)
+#endif
```
milseman, line 18: I'm not that familiar with manifests. Why is conditional compilation needed here?

milseman (commented):
LGTM. @compnerd, what do you think?

Package.swift
stevapple: 🤔 AFAIK this is used to comfort SwiftPM 5.3 because it doesn't recognize Windows.  This workaround is introduced by @compnerd for other projects and I just use it here.

compnerd: I don't think that this would be needed given the comment below.

Package.swift
```diff
@@ -12,11 +12,19 @@
...
+
+let targets: [Target] = [
   .target(
     name: "SystemPackage",
     dependencies: ["CSystem"],
     path: "Sources/System",
+    cSettings: [
+      .define("_CRT_SECURE_NO_WARNINGS", .when(platforms: windowsPlatform)),
```
compnerd, line 26: We can do this unconditionally, it won't impact other targets.  Alternatively, we _could_ clean up the cases to use the Microsoft "extensions" (technically I believe that they are part of Annex G, so they _should_ be portable).

stevapple: Sounds reasonable. Fixed

milseman (approved):
LGTM

milseman:
@swift-ci please test

compnerd (approved):

milseman:
@swift-ci please test

karwa:
Can confirm that the Windows build is broken, and that this patch fixes it. This is a blocking issue for me; is there anything you need before this can be merged @milseman? Also, once merged, it would be great if you could tag a new release so libraries which depend on `swift-system` can pick up the fix. 

There are still some deprecation warnings (included below), which should probably be addressed at some point, but at least this patch gets things building again.

```
C:\Users\User\Documents\Code\swift-url\.build\checkouts\swift-system\Sources\System\Internals\Syscalls.swift:82:10: warning: 'lseek' is deprecated: The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _lseek. See online help for details.
  return lseek(fd, off, whence)
         ^
C:\Users\User\Documents\Code\swift-url\.build\checkouts\swift-system\Sources\System\Internals\Syscalls.swift:109:10: warning: 'dup' is deprecated: The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _dup. See online help for details.
  return dup(fd)
         ^
C:\Users\User\Documents\Code\swift-url\.build\checkouts\swift-system\Sources\System\Internals\Syscalls.swift:116:10: warning: 'dup2' is deprecated: The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _dup2. See online help for details.
  return dup2(fd, fd2)
         ^
```

milseman:
My bad!

milseman:
@karwa I made a draft 0.0.3 release here: https://github.com/apple/swift-system/releases/tag/untagged-cb0bff1f58c393ba2621

Are you able to try it out before I publish it? The release process is still a little awkward as we're preparing for a proper 1.0 soon.

stevapple:
@milseman Although this is off-topic here, could I expect using both POSIX and Windows path on either platform as for 1.0 release? This useful for cross-platform things or simply using POSIX path style on Windows (eg. in-memory file system in SwiftPM).

milseman:
Unfortunately not. While we can support that technology internally, we don't have a good way to surface that as API yet. I'm hoping that task-local values gives us a way of establishing a contextual platform path type. We can also explore a FilePathProtocol and have a family of various concrete conformers for explicitly picking one over the other, but there are many many design axis there and that would take more consideration than we can do before 1.0.

I'd be interested in your thoughts on how you'd want to use such an ability though.

karwa:
@milseman Apologies for the delay. I'm not sure how to get the link you posted to work, but I switched the dependency to the `main` branch, and it does work.

The README currently suggests using `upToNextMinor` for source-stability reasons, and anybody who is following that advice will not automatically get `1.0`, so I think it may still make sense to tag a `0.0.3` release regardless.

That being said, probably all packages which use `swift-system` as a dependency should release new versions which bump their requirements to `1.0` anyway. So I don't know ⚖️

milseman:
Published 0.0.3

PR #56, opened 2021-07-09
[DRAFT][SKETCH] Ideas for stat.h interface

rauhul:
- Adds wrapper functions for all _posix_ functions in darwin's `stat.h` header.
- Non-posix `stat.h` functions (`<fn>_np`) are out of scope of this sketch.
- `fcntrl` functions are flags are out of scope of this sketch.

Needs work:
- Function names
- API for `umask`
- Documentation
- Decide if `FileModeOther` should exist
- Tests


milseman:
I want to make sure we can make (even if slow) progress on this. @rauhul, have you thought any more about this?

rauhul:
@milseman I think we need a good story about `fcntl` flags and if want to move the non-permissions bits of FilePermissions out into a different type. In the sketch as is, these are called FileModeOther as a placeholder. I think two other open areas are `TimeSpecification` and function naming.

rauhul:
@glessard would you be interested in reviving this PR?

Issue #57, opened 2021-07-12
Add `FileDescriptor.openPipe`


GeorgeLyon:
> I had mentally lumped that under process types and calls, but it's not actually blocked on any design work there. We can add this anytime and might make a good starter task.

_Originally posted by @milseman in https://github.com/apple/swift-system/issues/16#issuecomment-877686504_

One thing that isn't completely clear to me is whether or not it would be useful to introduce a `Pipe` type (or `FileDescriptor.Pipe` so as not to cloud the top level namespace). I'm not sure how big of a deal this is, but this could prevent bugs like `let (write, read) = try FileDescriptor.openPipe()`, as well as allow us to implement `close` and `closeAfter` for both file descriptors at once. (Not sure how this affects Windows, @compnerd)

You are correct, this shouldn't be too difficult, at least on the Linux/macOS side of things. I'd be happy to take a pass if we agree on whether or not to include `Pipe`.

Issue #54, opened 2021-07-07
Should `FileDescriptor.closeAfter` provide a `FileDescriptor` to the invoked closure?


GeorgeLyon:
I've been wanting to write something like

```swift
let foo = try FileDescriptor
  .open("my-file", .readOnly)
  .closeAfter { fileDescriptor in
    /// do stuff with fileDescriptor
  }
```

PR #58, opened 2021-07-21
Implement `FileDescriptor.Pipe()`

GeorgeLyon:
See #57 for discussion.


milseman:
@lorentey I haven't looked into pipes too deeply, but this seems like the right direction. There are some minor details I'm not sure about, such as an initializer backed by a syscall vs a static create method (just haven't thought it through), etc. WDYT?

GeorgeLyon:
I don't have a strong preference on this, I was just mirroring how `FileDescriptor` is created by `open`.

lorentey (changes requested):
This would be very welcome addition! 👍 

I think I'd prefer if we went with a simple `pipe()` method, unless there is a good reason to have the `Pipe` type that I'm not seeing.

Sources/System/FileHelpers.swift
```diff
@@ -123,3 +123,34 @@ extension FileDescriptor {
...
+  /// - Parameter body: The closure to run.
+  ///   If the closure throws an error,
+  ///   this method closes both ends of this pipe before it rethrows that error.
+  ///
+  /// - Returns: The value returned by the closure.
+  ///
+  /// If `body` throws an error
+  /// or an error occurs while closing the file descriptor,
+  /// this method rethrows that error.
+  public func closeAfter<R>(_ body: () throws -> R) throws -> R {
```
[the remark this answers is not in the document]
GeorgeLyon: There are two scenarios where this is useful:
1. You are forking off **two** child processes and simply connecting them with a pipe
2. In an async/await world, you will be able to write or read some data entirely within the `closeAfter` block

GeorgeLyon:
I've addressed some of @lorentey's feedback. I still think a separate structure is warranted for the reasons listed here: https://github.com/apple/swift-system/pull/58#discussion_r683740434, but let me know if you disagree.

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,52 @@ extension FileDescriptor {
   public func dup2() throws -> FileDescriptor {
     fatalError("Not implemented")
   }
+  
+  /// A pair of `FileDescriptor` values represening a pipe.
+  ///
+  /// You are responsible for managing the lifetime and validity
+  /// of the `read` and `write` `FileDescriptor` values,
+  /// which can be closed independently or via helper methods on `Pipe`, but not both.
+  public struct Pipe {
```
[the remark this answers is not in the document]
milseman: Mentioned briefly in https://github.com/apple/swift-system/issues/57. I think the intention is to help avoid mixing up the read and write ends of the pipe, as tuple labels are not strongly enforced for destructuring assignment.

There's a point at which it's worth turning a labeled tuple into a proper nominal type, but I'm not sure where on the spectrum pipe falls. If it is usually the case that pipes are immediately split apart, then I think a tuple is fine. On the other hand, are both ends of a pipe typically closed together?

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,52 @@ extension FileDescriptor {
...
+    internal static func _create(
+      retryOnInterrupt: Bool
+    ) -> Result<Pipe, Errno> {
+      var fds: [Int32] = [-1, -1]
+      return valueOrErrno(retryOnInterrupt: retryOnInterrupt) {
+        system_pipe(&fds)
+      }.map { _ in Pipe(fds: fds) }
+    }
+    
+    private init(fds: [Int32]) {
```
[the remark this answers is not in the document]
milseman: I know the stdlib proper has conventions along those lines, much of it due to historical limitations. If the init doesn't appear in the ABI at all, then I'm not sure what value that convention brings for Swift System.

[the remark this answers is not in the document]
GeorgeLyon: I haven't contributed much to either, but I've changed this to `_input:output:` to match @lorentey's recommendations below. Let me know if I should change it to `input:output:`.

Sources/System/FileOperations.swift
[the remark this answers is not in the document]
GeorgeLyon: One other consideration: if we change `FileDescriptor.closeAfter` to pass the file descriptor to the provided closure (as suggested in #54) then having similar behavior for `Pipe` would be nice (and require a dedicated `struct`). The main use case I see for closing both ends of a pipe at once is mentioned [here](https://github.com/apple/swift-system/pull/58#discussion_r683740434), namely forking off two processes you want connected with a pipe. This is important for shell-like and orchestration use cases.

Sources/System/FileHelpers.swift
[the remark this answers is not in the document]
milseman: For scenario 1, are you handing off one FD to a child and another to the other child, or do both get the full pipe?

For scenario 2, could you give more context?

If we are making a proper pipe type, does it make sense to have read/write operations on it that would hook up to the appropriate end?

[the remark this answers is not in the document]
GeorgeLyon: > For scenario 1, are you handing off one FD to a child and another to the other child, or do both get the full pipe?

One to each, the simplest example of this is implementing something like `$ echo foo | sed 's/foo/bar/'` in Swift. The naive way to solve this without `closeAfter` is wrong:
```
let (input, output) = pipe
try input.closeAfter {
  try launchChildProcess(input: input)
}
try output.closeAfter {
  try launchChildProcess(output: output)
}
```
The issue here is that if the first `launchChildProcess` fails, `output` will not be closed. If we do `closeAfter` on the pipe, this behaves correctly.

> For scenario 2, could you give more context?

With #54, I'm imagining something like:
```
let output: String = try await Pipe().closeAfter { pipe in
  try launchChildProcess(output: pipe.input)
  return try await pipe.output.readToEOF()
}
```

> If we are making a proper pipe type, does it make sense to have read/write operations on it that would hook up to the appropriate end?

Perhaps? I don't see a huge benefit to `Pipe.read` over `Pipe.output.read`, though perhaps it will discourage less sensible usage like `Pipe.output.write`. Either way, this can be added as a subsequent improvement.

**ALSO** My first code example is actually wrong, and launching child processes should look something like this `launchChildProcess(input: output)`. Once we actually get around to implementing something along the lines of `posix_spawn` we can do the same thing `Process` does and accept either a `FileDescriptor` or a `Pipe` and in the case that a `Pipe` is provided, the process-launching code will select the correct end of the pipe to use depending on whether it is passed for stdin/stdout or stderr.

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,52 @@ extension FileDescriptor {
...
   }
+  
+  /// A pair of `FileDescriptor` values represening a pipe.
+  ///
+  /// You are responsible for managing the lifetime and validity
+  /// of the `read` and `write` `FileDescriptor` values,
+  /// which can be closed independently or via helper methods on `Pipe`, but not both.
+  public struct Pipe {
+    /// The file descriptor for reading from this pipe
+    public let fileDescriptorForReading: FileDescriptor
```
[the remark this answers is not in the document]
milseman: But is it "input" or "output"? The end for reading gives you the content of `write` calls, so it's the output of a pipe, or am I mixing this up? Naming is very important here...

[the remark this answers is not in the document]
GeorgeLyon: It may be very Objective-C of me, but this is why I initially went with the `fileDescriptorForReading` (this mirrors prior art in `Foundation.Pipe`). I tend to prefer more verbose names, except in cases where there is a sufficiently descriptive term-of-art. I'm by no means an expert in POSIX, so I'll defer to you.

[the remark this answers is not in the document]
GeorgeLyon: I reverted this back to use the more verbose names, which I prefer more.

karwa:
I think the "fileDescriptor" part of the name is superfluous.

What if we went further with the pipe terminology and called them "inlet" and "outlet"? i.e. the descriptor used to push data in to the pipe and the descriptor where data emerges from?

milseman:
The more I think about this PR, the more I like having a nominal type (inside `FileDescriptor`) for the pipe with well-named members. Just as `FileDescriptor` is clearly modeling the OS concept of an entry in a descriptor table and shouldn't be confusable with a theoretical future `File` moveonly struct, I feel that `FileDescriptor.Pipe` won't be confused with some theoretical future type used with a `Process` actor or anything like that.

I really like @karwa 's suggestion's (`inlet` / `outlet`) as thematically consistent and clear, assuming there's not a better "standard" Swift name for the concept.

GeorgeLyon:
I changed it to inlet/outlet, great suggestion @karwa!

Sources/System/FileOperations.swift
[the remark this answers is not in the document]
lorentey: > I know the stdlib proper has conventions along those lines, much of it due to historical limitations. If the init doesn't appear in the ABI at all, then I'm not sure what value that convention brings for Swift System.

The primary benefit of underscoring non-public names in my view is consistency and simplicity. I much prefer rules that are easy to explain, especially to folks who aren't necessarily well-versed in the intricacies of Swift's ABI features. I think it's reasonable to expect that people know what a public declaration is -- I don't think it's reasonable to expect that the average Swift developer would know exactly what declarations have an ABI impact. (Simply because it's so unusual to work on an ABI stable library.) So I think it's better to simply say that the policy is that the name of anything that's not declared `public` must carry an underscore. (The need to add an underscore also helps eliminate never-ending naming discussions for these internals, just because it makes these names "imperfect", by design, lowering stakes.)

Underscoring everything that's not public also helps prevent mistakes when previously private/internal declarations get updated to `@usableFromInline` or `public`. However, this is only a minor benefit.

Sources/System/FileOperations.swift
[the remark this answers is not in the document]
lorentey: The thing is, mixing up the read and write ends of the pipe is a fairy cheap mistake to make, as mistakes go. Reading from or writing to the wrong end will immediately throw an error. 

I do agree it would be nice to prevent such mixups at compile time, but going with a struct makes structural decomposition a little more difficult, and (more importantly) it gives the (in my opinion) false impression that there is an API-level abstraction called `Pipe` that comes with its own operations. There are no operations on pipes -- there are only operations on its independent end points.

Sources/System/FileOperations.swift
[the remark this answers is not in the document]
lorentey: The terms `input` and `output` in my suggestion reflect the naming convention as defined by `FileDescriptor.standardInput` and `.standardOutput` -- an "input filedescriptor" is one that you can read from; an "output fd" is one you can write to.

```swift
let (input, output) = FileDescriptor.pipe()
defer { 
  try! input.close() // Closing a valid pipe never errors out.
  try! output.close()
}
try launchChildProcess(input: .standardInput, output: output)
try launchChildProcess(input: input, output: .standardOutput)
```

lorentey:
Naming them `inlet` and `outlet` is an attractive idea, but I think they just make things worse:

```swift
let (outlet, inlet) = FileDescriptor.pipe()
defer { 
  try! outlet.close() // Closing a valid pipe never errors out.
  try! inlet.close()
}
try launchChildProcess(input: .standardInput, output: inlet) // ???
try launchChildProcess(input: outlet, output: .standardOutput) // ???
```

Repeating https://github.com/apple/swift-system/pull/58#discussion_r699677231:

The terms `input` and `output` in my suggestion reflect the naming convention as defined by `FileDescriptor.standardInput` and `.standardOutput` -- an "input filedescriptor" is one that you can read from; an "output fd" is one you can write to.

```swift
let (input, output) = FileDescriptor.pipe()
defer { 
  try! input.close() // Closing a valid pipe never errors out.
  try! output.close()
}
try launchChildProcess(input: .standardInput, output: output)
try launchChildProcess(input: input, output: .standardOutput)
```

karwa:
It's not so bad if you use them from the pipe object/tuple:

```swift
let pipe = try FileDescriptor.Pipe()
defer { 
  try! pipe.close() // Closing a valid pipe never errors out.
}
try launchChildProcess(input: .standardInput, output: pipe.inlet) // Oh, ok - outputting to the pipe
try launchChildProcess(input: pipe.outlet, output: .standardOutput) // Oh, ok - reading from the pipe
```

But yeah, if you immediately destructure the result in to 2 file descriptors in the calling function, those names might not make sense and you can choose to bind them to other names in your destructuring.

Sources/System/FileHelpers.swift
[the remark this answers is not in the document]
lorentey: > Once we actually get around to implementing something along the lines of `posix_spawn` we can do the same thing `Process` does and accept either a `FileDescriptor` or a `Pipe` and in the case that a `Pipe` is provided, the process-launching code will select the correct end of the pipe to use depending on whether it is passed for stdin/stdout or stderr. 

This is a good argument.

`Process` adds a large number of high-level assumptions that aren't built into `posix_spawn` -- for instance, `posix_spawn` does not assume that fd 0 must be an input file descriptor. (It most often is, but it's just a convention -- in its direct syscall wrappers, System probably should not introduce assumptions that would prevent its use in niche cases.) Without that assumption, a helper that takes a pipe for fd 0 is less obviously desirable.

But this is a nitpicky argument -- whatever wrapper we create for `posix_spawn_file_actions_t`, it seems desirable to have a set of convenience methods to set up conventional stdin/stdout/stderr channels.

lorentey:
I don't know, `pipe.inlet` doesn't look like a huge improvement to me. For an argument labeled `output:`, my default instinct would be to type `pipe.` then select whatever matches the label best from the completion list.

Going from `output` to `inlet` requires a bit of mental gymnastics. It's clever and correct, but it's not helpful.

GeorgeLyon:
To further improve on @karwa's example, you can just pass the pipe:
```
try launchChildProcess(input: .standardInput, output: pipe)
try launchChildProcess(input: pipe, output: .standardOutput)
```
And `launchChildProcess` would know to use `inlet` for output and `outlet` for input (we can accomplish this with overloads or protocols). This way there is not danger of making a mistake. 

---

Even without that, `inlet` may not be _perfect_ but it is certainly better in that case than `input` since it would be much easier to type `input: pipe.input` and miss the error. Having it be `inlet` at least gives you a bit more opportunity to do a double-take.

lorentey:
> Even without that, `inlet` may not be perfect but it is certainly better in that case than `input` since it would be much easier to type `input: pipe.input` and miss the error.

What error? 😊

My entire point is that passing `pipe.input` for the `input:` argument would be correct: the labels align perfectly with the names for pipe components, as well as with the existing `standardInput`/`standardOutput` terminology. (`pipe.input` is what @karwa calls the "outlet".)

```swift
// This looks obviously correct, and is actually correct. It's also eminently consistent.
try spawnChildProcess(input: .standardInput, output: pipe.output)
try spawnChildProcess(input: pipe.input, output: .standardOutput)

// This looks incorrect, but is actually correct. It's clever, but tiresome. I'll get it wrong every single time.
try spawnChildProcess(input: .standardInput, output: pipe.inlet)
try spawnChildProcess(input: pipe.outlet, output: .standardOutput)
```

This also rhymes with how `pipe()` *happens to* line up the indices of its result array with the "0 means stdin, 1 means stdout" convention.

lorentey:
> It seems we have two options:

I realize it gets a little worse, as we'd need `spawnChildProcess` to close the unused half of the pipes it is given at the start of the new process. This eliminates option 2, and upgrades the overloads in option 1 from being just a `pipe` → `pipe.input`/`output` shortcut to something a little more involved.

Practically, I don't think `spawnChildProcess` would be the right way to expose pipes on a high-level. A more likely high-level approach would be to create some sort of declarative API for hooking processes into a pipeline of connected stdout/stdins:

```swift
let pipeline: Pipe = [
  Command("cat", "swift-system/README.md")
  Command("tr", "[:space:]", "\n")
  Command("sort")
  Command("uniq", "-c")
  Command("sort", "-nr")
  Command("head", "-10")
])
try await try pipeline.run()
// Prints to stdout:
  96 
   9 the
   7 for
   6 to
   6 and
   6 a
   5 is
   5 System
   5 =
   5 ##
```

I believe such a facility would be very desirable, but (in my view) it belongs in a package on top of System, rather than embedded in it. Whether `pipe()` returned a named type or a tuple wouldn't matter one bit -- that is not where the difficulties lie.

---

On the other hand, having a named `FileDescriptor.Pipe` type would allow us to add a finite, small set of convenience methods to System's posix_spawn wrapper, though:

```swift
let file = try FileDescriptor.open("/tmp/foo.txt", .readOnly)
defer { try! file.close() }
let pipe = try FileDescriptor.pipe()
defer { try! pipe.close() } // shortcut!

var actions: Spawn.Actions = []

actions.setStandardInput(to: file)
// Expands to:
  actions.close(.standardInput)
  actions.duplicate(file, as: .standardInput)
  actions.close(file)

actions.setStandardOutput(to: pipe)
// Expands to:
  actions.close(pipe.input)
  actions.close(.standardOutput)
  actions.duplicate(pipe.output, as: .standardOutput) // Note how the terminology lines up again!
  actions.close(pipe.output)

let pid = try ProcessID.spawn("/usr/bin/sort", actions, ["/usr/bin/sort", "-r"])
```

I don't think that System *must* provide such action shortcuts -- but if y'all think this is something worth spending a named type on, I will not continue arguing against it, either.  At the end of the day, it is a relatively minor thing. 😅

The dup2 example action above underscores the importance of consistent naming though -- I think `actions.duplicate(pipe.inlet, as: .standardOutput)` wouldn't work nearly as well.

GeorgeLyon:
I'm swayed by @lorentey's argument. I have replaced the `Pipe` `struct` with a `typealias` which should not impact the ABI, but give a bit of sugar for those that want it.
I agree that writing to the wrong end of a pipe is a "cheap" mistake, and the various affordances we have discussed (automatically selecting the correct end of a pipe, `closeAfter`) can be implemented in a higher level API.

Sources/System/FileOperations.swift
```diff
@@ -373,52 +373,34 @@ extension FileDescriptor {
...
-      }
-    }
-    
-    private init(_fileDescriptorForReading: Int32, fileDescriptorForWriting: Int32) {
-      self.outlet = FileDescriptor(rawValue: _fileDescriptorForReading)
-      self.inlet = FileDescriptor(rawValue: fileDescriptorForWriting)
-    }
-  }
+  /// of the `output` and `input` `FileDescriptor` values.
+  public typealias Pipe = (output: FileDescriptor, input: FileDescriptor)
```
lorentey, line 379: The labels are swapped here -- `fds[0]` is the input file descriptor, `fds[1]` is the output; they are 

```suggestion
  public typealias Pipe = (input: FileDescriptor, output: FileDescriptor)
```

Sources/System/FileOperations.swift
```diff
@@ -373,52 +373,34 @@ extension FileDescriptor {
...
+  ///      Pass `false` to try only once and throw an error upon interruption.
+  /// - Returns: The new file descriptor.
+  ///
+  /// The corresponding C function is `pipe`.
   @_alwaysEmitIntoClient
-  @available(*, unavailable, renamed: "FileDescriptor.Pipe()")
-  public func pipe() throws -> FileDescriptor {
-    fatalError("Not implemented")
+  // @available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
+  public static func pipe(retryOnInterrupt: Bool = true) throws -> Pipe {
```
lorentey, line 392: `pipe()` is not an interruptible syscall; why do we need `retryOnInterrupt` here?

GeorgeLyon: Sorry this was copy pasta

Sources/System/FileOperations.swift
GeorgeLyon: Fixed

milseman (commented):
This naming is backwards from the pipe's perspective and is confusing. I get the stdin/stdout analogy, but I do not believe it applies in this case (or, rather, it can just as easily apply in reverse).

If a "thing" (i.e. an encapsulation) has a member named input or output, it's the thing's input/output. It's not the input/output of whatever is holding the thing. Input and output are terms relative to some frame of reference and the concepts flip when viewed from the inside vs the outside.

Stdin/stdout are the input and output of the currently executing process, that's why they're globals. That is, globals are "local" members inside of the current process. Inside the process, we read from the input and write to the output, hence the names. Outside that process, we communicate with a process by writing to its input and/or reading from its output.

Even if it's just a typealias, giving it a name like `Pipe` is part of the source stability story. Giving it a name (even if via typealias) reifies it into a "thing" that has members.

If we're going with just a tuple, then the labels are less salient and can be ignored by someone familiar with the ordering. Otherwise they provide good documentation and I would prefer a clearer name. "Input/output" is relative to some frame of reference. `man 2 pipe` terminology of "read end" and "write end" is absolute, as is plumbing terminology of "inlet" and "outlet". `(readEnd: FD, writeEnd: FD)`, `(outlet: FD, inlet: FD)`, etc., read better to me than `(input: FD, output: FD)` while lacking a frame of reference.

### Trouble with Tuples

As for tuples, I've usually regretted any time that I've settled on a typealias for a tuple rather than a real nominal type. It's unfortunate that real nominal types are such a pain to write in Swift.

An API pattern I've been playing around with elsewhere for such nominalized types is to have a `var destructure: (fieldA: A, fieldB: B, ...)`, which gives you the tuple version if you really want/need it (e.g. switch expressions). This probably isn't the time to formalize this pattern in System.

I want to have a more comprehensive treatment of the file descriptor type hierarchy at some point and I don't mean to balloon this PR into that. The OS concept of file descriptor has been ad-hoc overloaded and extended over the decades (always easier to awkwardly extend an existing construct than invent a new one), but we want to provide stronger/better types when we can deliver extra value in doing so. I don't think it's too crazy to make a `RawRepresentable` struct for pipes and a `FileDescriptorProtocol` or some such family of related protocols. We currently have (albeit in sketches/draft PRs) stronger types such as `SocketDescriptor` and `KernelQueue`. These have different subsets of operations that make sense on them and each has additional operations available.

IIUC, `pipe.write(...)` and `pipe.read(...)` have an obvious behavior of forwarding to the appropriate end. Closing a pipe is weird though because (IIUC) you're likely to close the write end, but the read end is still open for reading anything buffered until it gets a zero (meaning EOF). A nominal type also lets us say  `FileDescriptor.Pipe.open()` instead of just `FileDescriptor.pipe()`, which helps drive some of the intuition that the result should be closed (just like `FileDescriptor.open()`).

### For this PR

I'm inclined to drop the type alias for now. That at least saves the name `FileDescriptor.Pipe` for future nominalization. Or, we could just nominalize it now...

milseman:
@GeorgeLyon sorry for the long back-and-forth. It's frustrating that such a simple API sparks so much debate and deferred API design tradeoffs. I think @lorentey and I have settled on:

1) This not being a big enough deal to agonize over, so let's get something decent in place and merge.
2) Avoid the tuple type alias. Type aliased tuples are just not quite production-ready in Swift.
3) Let's call it `(readEnd: FD, writeEnd: FD)`. A pipe has 2 ends, one for reading and the other for writing, so let's not add a layer of terminology indirection via "outlet/inlet" nor assume an inside/outside polarity with "input/output".

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,37 @@ extension FileDescriptor {
...
     fatalError("Not implemented")
   }
+  
+  /// A pair of `FileDescriptor` values represening a pipe.
+  ///
+  /// A pipe enables data written to `output` to be read from `input`.
+  /// You are responsible for managing the lifetime and validity
+  /// of the `input` and `output` `FileDescriptor` values.
+  public typealias Pipe = (input: FileDescriptor, output: FileDescriptor)
+
```
milseman, line 380:
```suggestion
```

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,37 @@ extension FileDescriptor {
...
   }
+  
+  /// A pair of `FileDescriptor` values represening a pipe.
+  ///
+  /// A pipe enables data written to `output` to be read from `input`.
+  /// You are responsible for managing the lifetime and validity
+  /// of the `input` and `output` `FileDescriptor` values.
+  public typealias Pipe = (input: FileDescriptor, output: FileDescriptor)
+  
+  /// Create a pipe, a uniderctional data channel which can be used for interprocess communication.
```
milseman, line 381:
```suggestion
  /// Create a pipe, a unidirectional data channel which can be used for interprocess communication.
```

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,37 @@ extension FileDescriptor {
...
+  ///
+  /// - Parameters:
+  ///   - retryOnInterrupt: Whether to retry the write operation
+  ///      if it throws ``Errno/interrupted``. The default is `true`.
+  ///      Pass `false` to try only once and throw an error upon interruption.
```
milseman, line 386:
```suggestion
```

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,37 @@ extension FileDescriptor {
...
+  /// - Returns: The new file descriptor.
```
milseman, line 387:
```suggestion
  /// - Returns: The pair of file descriptors.
```
```suggestion
  /// - Returns: The new file descriptor.
```

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,37 @@ extension FileDescriptor {
...
+  ///
+  /// The corresponding C function is `pipe`.
+  @_alwaysEmitIntoClient
+  // @available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
+  public static func pipe() throws -> Pipe {
```
milseman, line 392:
```suggestion
  public static func pipe() throws -> (readEnd: FileDescriptor, writeEnd: FileDescriptor) {
```

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,37 @@ extension FileDescriptor {
...
+    try _pipe().get()
+  }
+  
+  @usableFromInline
+  internal static func _pipe() -> Result<Pipe, Errno> {
```
milseman, line 397:
```suggestion
  internal static func _pipe() -> Result<(readEnd: FileDescriptor, writeEnd: FileDescriptor), Errno> {
```

Also probably needs availability...

Tests/SystemTests/FileOperationsTest.swift
```diff
@@ -89,6 +89,25 @@ final class FileOperationsTest: XCTestCase {
   func testHelpers() {
     // TODO: Test writeAll, writeAll(toAbsoluteOffset), closeAfter
   }
+  
+  func testAdHocPipe() throws {
+    // Ad-hoc test testing `Pipe` functionality.
+    // We cannot test `Pipe` using `MockTestCase` because it calls `pipe` with a pointer to an array local to the `Pipe`, the address of which we do not know prior to invoking `Pipe`.
```
milseman, line 95: We probably do (not gonna hold up this PR for it though) want to allow for mock testing of such stack local pointers. We'd probably have a variant that we'd pass in an array and it would compare the contents.

GeorgeLyon:
No worries! This matters so it is design, not bike-shedding and design is important :)

I like the "readEnd", "writeEnd" nomenclature. For now I just blindly accepted the review suggestions, but I can do a second pass to make sure everything builds, etc. I'm not sure how you all handle formatting/CI etc so let me know if there is something else I can do (or if it is simpler feel free to just commandeer this PR).

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,26 @@ extension FileDescriptor {
...
+  
+  @usableFromInline
+  internal static func _pipe() -> Result<(readEnd: FileDescriptor, writeEnd: FileDescriptor), Errno> {
+    var fds: (Int32, Int32) = (-1, -1)
+    return withUnsafeMutableBytes(of: &fds) { bytes in
+      let fds = bytes.bindMemory(to: Int32.self)
+      return valueOrErrno(retryOnInterrupt: false) {
+        system_pipe(fds.baseAddress!)
+      }.map { _ in (FileDescriptor(rawValue: fds[0]), FileDescriptor(rawValue: fds[1])) }
+    }
```
milseman, line 393: You might be able to say `var fds: Array<CInt> = [-1, -1] ` and use the implicit array-to-pointer conversion (thought I don't recall if it's mutable). Otherwise, you can probably use `withUnsafeMutablePointer` to avoid the extra bind-memory step.

milseman:
@swift-ci please test

Sources/System/FileOperations.swift
GeorgeLyon: @lorentey suggested I use the tuple form for a stronger guarantee that those values would end up on the stack. I haven't checked but I believe `withUnsafeMutablePointer` will still require rebinding the type (I think this was a signedness issue), but would also require manually passing the count (1), whereas `withUnsafeMutableBytes` takes this information from the buffer pointer.

compnerd: Please do make sure that this builds on windows - I'm almost certain that this will break the Windows builds.  Also note that Windows should transact in `HANDLE`s, aka `void *`, so this is going to truncate at least.  See [`CreatePipe`](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-createpipe).

lorentey: I believe it's fine to simply surround these additions with `#if !os(Windows)` for now.

Tests/SystemTests/FileOperationsTest.swift
```diff
@@ -89,6 +89,25 @@ final class FileOperationsTest: XCTestCase {
...
+    let pipe = try FileDescriptor.pipe()
+    try pipe.input.closeAfter {
+      try pipe.output.closeAfter {
+        var abc = "abc"
+        try abc.withUTF8 {
+          _ = try pipe.output.write(UnsafeRawBufferPointer($0))
+        }
+        let readLen = 3
+        let readBytes = try Array<UInt8>(unsafeUninitializedCapacity: readLen) { buf, count in
+          count = try pipe.input.read(into: UnsafeMutableRawBufferPointer(buf))
```
milseman, line 105:
```suggestion
          count = try pipe.readEnd.read(into: UnsafeMutableRawBufferPointer(buf))
```

Tests/SystemTests/FileOperationsTest.swift
```diff
@@ -89,6 +89,25 @@ final class FileOperationsTest: XCTestCase {
   func testHelpers() {
     // TODO: Test writeAll, writeAll(toAbsoluteOffset), closeAfter
   }
+  
+  func testAdHocPipe() throws {
+    // Ad-hoc test testing `Pipe` functionality.
+    // We cannot test `Pipe` using `MockTestCase` because it calls `pipe` with a pointer to an array local to the `Pipe`, the address of which we do not know prior to invoking `Pipe`.
+    let pipe = try FileDescriptor.pipe()
+    try pipe.input.closeAfter {
+      try pipe.output.closeAfter {
```
milseman, line 98:
```suggestion
      try pipe.writeEnd.closeAfter {
```

milseman:
@swift-ci please test

Sources/System/FileOperations.swift
milseman: This is looking great! @atrick / @glessard , what's the lesser of evils for the pointer binding stuff?

milseman (commented):
I'm ready to merge as soon as the pointer binding thing is figured out.

Sources/System/FileOperations.swift
karwa: The pointer to a homogenous tuple is also bound to its element type, so you can use `assumingMemoryBound`.

_(I hope. I use this quite a lot)_

glessard: As written, this case indeed calls for `assumingMemoryBound`.  
The memory is _already_ bound to `Int32`, so `bindMemory` doesn't add information.

This being said, we unfortunately don't have `assumingMemoryBound` on the `RawBufferPointer` types at this point. We also gain nothing from using the Buffers in the case: we must rely in the C call to be well behaved instead of having any sort of bounds checking. Given that, I suggest this for lines 389-393:
```
    return withUnsafeMutablePointer(to: &fds) { pointer in
      valueOrErrno(retryOnInterrupt: false) {
        system_pipe(UnsafeMutableRawPointer(pointer).assumingMemoryBound(to: Int32.self))
      }
    }.map { _ in (FileDescriptor(rawValue: fds.0), FileDescriptor(rawValue: fds.1)) }
```

atrick: @karl is correct, and @glessard 's code edit looks good.
Although the original code was harmless, it doesn't really make sense to bind memory that belongs to a variable.

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,28 @@ extension FileDescriptor {
...
+  }
+  
+  @usableFromInline
+  internal static func _pipe() -> Result<(readEnd: FileDescriptor, writeEnd: FileDescriptor), Errno> {
+    var fds: (Int32, Int32) = (-1, -1)
+    return withUnsafeMutableBytes(of: &fds) { bytes in
+      let fds = bytes.bindMemory(to: Int32.self)
+      return valueOrErrno(retryOnInterrupt: false) {
+        system_pipe(fds.baseAddress!)
+      }.map { _ in (FileDescriptor(rawValue: fds[0]), FileDescriptor(rawValue: fds[1])) }
```
glessard, line 393: (restating my suggestion in the form of a suggestion.)

```suggestion
    return withUnsafeMutablePointer(to: &fds) { pointer in
      valueOrErrno(retryOnInterrupt: false) {
        system_pipe(UnsafeMutableRawPointer(pointer).assumingMemoryBound(to: Int32.self))
      }
    }.map { _ in (.init(rawValue: fds.0), .init(rawValue: fds.1)) }
```

Sources/System/FileOperations.swift
```diff
@@ -370,4 +370,28 @@ extension FileDescriptor {
...
+  
+  @usableFromInline
+  internal static func _pipe() -> Result<(readEnd: FileDescriptor, writeEnd: FileDescriptor), Errno> {
+    var fds: (Int32, Int32) = (-1, -1)
+    return withUnsafeMutablePointer(to: &fds) { pointer in
+      valueOrErrno(retryOnInterrupt: false) {
+        system_pipe(UnsafeMutableRawPointer(pointer).assumingMemoryBound(to: Int32.self))
+      }
+    }.map { _ in (.init(rawValue: fds.0), .init(rawValue: fds.1)) }
+    }
```
milseman, line 394:
```suggestion
```

milseman:
@swift-ci please test

lorentey (approved):

GeorgeLyon:
I do not, but thank you for the thought :) Happy to have this landed!

lin7sh:
Late for the party, curious about is this optimised for Darwin's mach port implementation?

Sources/System/FileOperations.swift
GeorgeLyon: Is there a document detailing what "binding memory" actually does? I had thought this was just managing compile-time information so I'm not 100% clear on the difference between "assuming" memory bound and actually binding memory.

glessard: There is some scattered documentation, including the doc-comments, but the largest chunk is in the `RawPointer` proposal: https://github.com/apple/swift-evolution/blob/main/proposals/0107-unsaferawpointer.md  
Experience has shown that those sources are insufficient.

@NevinBR (I think) came up with a nice description of binding for humans a few weeks ago: https://forums.swift.org/t/pitch-implicit-pointer-conversion-for-c-interoperability/51129/36.

In general, if you're reminding the compiler of type information it should already have known, but had been obscured for some reason, then `assumingMemoryBound` is the thing to reach for (as we did here). If you are telling the compiler new information, then reach for `bindMemory` or `withMemoryRebound`.

lorentey:
> Late for the party, curious about is this optimised for Darwin's mach port implementation?

I'm sorry, I don't understand this question. This PR adds a lightweight Swift wrapper around the classic UNIX `pipe` system call. Pipes are a primitive IPC mechanism that predate Mach ports.

The PR doesn't change how `pipe(2)` works (that's not within Swift System's mandate) -- it merely exposes an idiomatic Swift interface for it that faithfully reflects its existing semantics.

glessard:
I believe that `pipe(2)` is implemented in the "open-source" part of Darwin's xnu kernel. You can probably find an answer at opensource.apple.com, where the periodic source releases are organized.

lin7sh:
thanks for information

PR #59, opened 2021-08-04
Fix SystemChar isLetter logic

rauhul:
- Fixes range expression in isLetter logic and adds test coverage.


rauhul:
@swift-ci please test

milseman (approved):
LGTM, thanks for the fix and test cases!

PR #66, opened 2021-10-30
Added Time types

colemancda:
Added Time, TimeInterval and Clock types for interfacing with System time.

### DIFF
--- a/Sources/System/Internals/CInterop.swift
+++ b/Sources/System/Internals/CInterop.swift
@@ -63,4 +63,30 @@ public enum CInterop {
   /// on API.
   public typealias PlatformUnicodeEncoding = UTF8
   #endif
+  
+  /// The C `clock_t` type
+  public typealias Clock = clock_t
+  
+  /// The C `clockid_t` type
+  public typealias ClockID = clockid_t
+      
+  /// The C `time_t` type
+  public typealias Time = time_t
+  
+  /// The C `tm` type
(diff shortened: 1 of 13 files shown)


Lukasa:
I suspect this is redundant with the proposed Time management types for the standard library: https://forums.swift.org/t/pitch-clock-instant-date-and-duration/52451

colemancda:
Well, we need some kind of wrapper to interface with these C APIs. It's my understanding that this library should wrap the C std lib APIs, like Foundation is a wrapper for CoreFoundation, ICU, cURL and libXML. If the functionality is part of the Swift standard library, thats great, but it should not depend on Glibc, Darwin, etc. `System` should provide Swift overlays for Darwin and Glibc C APIs.

Lukasa:
The Swift standard library should and does depend on libc when appropriate.

stevapple:
> It's my understanding that this library should wrap the C std lib APIs, like Foundation is a wrapper for CoreFoundation, ICU, cURL and libXML.

Nope.

`SwiftSystem` is a wrapper on, and a helper around **system programming** APIs, not the C standard library. The C standard library is already wrapping up mostly-used system APIs and it has a significantly higher abstraction level than what `System` provides.
Typically, we would encourage library designers to wrap up C APIs according to their use cases. There are plenty of existing best practices, especially server-side libraries.

PR #72, opened 2021-11-17
support arm64, for M1 Mac

DanboDuan:
it will fail if it is built on M1 Mac, so add arm64 for M1 Mac


cmake/modules/SwiftSupport.cmake
```diff
@@ -17,8 +17,12 @@ See https://swift.org/LICENSE.txt for license information
 function(get_swift_host_arch result_var_name)
   if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
     set("${result_var_name}" "x86_64" PARENT_SCOPE)
-  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
-    set("${result_var_name}" "aarch64" PARENT_SCOPE)
+  elseif ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "AArch64|aarch64|arm64")
```
compnerd, line 20: Hmm, I'm not sure if the `AArch64` entry is particularly useful currently, but the other two spellings are.  THis seems reasonable enough to me.  The only request is that we change the other packages similarly (dispatch, Foundation, XCTest, llbuild, Yams, swift-crypto, swift-driver, swift-tools-support-core, swift-package-manager, sourcekit-lsp).

compnerd (approved):

cmake/modules/SwiftSupport.cmake
DanboDuan: yes. this cmake code is just copied from swift-atomics/cmake/modules/SwiftSupport.cmake. or I can update it to only `arm64`  added

DanboDuan:
@compnerd hi, when could this pr be merged? or is there any thing I could update

PR #74, opened 2021-12-02
build: install CSystem into the system

compnerd:
CSystem is a leaky library dependency from System and must be
distributed.  Install the files and make the dependency public so that
it is emitted into the clients.

### DIFF
--- a/Sources/System/CMakeLists.txt
+++ b/Sources/System/CMakeLists.txt
@@ -34,7 +34,7 @@ target_sources(SystemPackage PRIVATE
   Internals/Mocking.swift
   Internals/Syscalls.swift
   Internals/WindowsSyscallAdapters.swift)
-target_link_libraries(SystemPackage PRIVATE
+target_link_libraries(SystemPackage PUBLIC
   CSystem)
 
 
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @milseman @tomerd

Sources/System/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ target_sources(SystemPackage PRIVATE
   Internals/Mocking.swift
   Internals/Syscalls.swift
   Internals/WindowsSyscallAdapters.swift)
-target_link_libraries(SystemPackage PRIVATE
+target_link_libraries(SystemPackage PUBLIC
```
milseman, line 37: Just for my own education, what does this do?

milseman (approved):

Sources/System/CMakeLists.txt
compnerd: This means that any user of `SystemPackage` will now transitively link against `CSystem`, which in this case propagates the include directives for the module definition (as it is an interface library).  Because `SystemPackage` publicly imports `CSystem` and forwards certain items as `@_emitIntoClient`, we need to ensure that all clients of `SystemPackage` can see `CSystem`.

tomerd:
@swift-ci test

PR #75, opened 2021-12-21
Unify availability annotations & introduce a script to manage them

lorentey:
Instead of spelling out availability incantations, contributors can now specify availability annotations using special comments that mention the swift-system version that will introduce the entry point:

```swift
/*System 0.0.2*/
public func greeting() -> String {
  "Hello"
}
```

The new script `Utilities/expand-availability.py` can then be used to augment these comments with the ABI-stable OS versions that introduced these:

```swift
/*System 0.0.2, @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)*/
public func greeting() -> String {
  "Hello"
}
```

(The script updates these expansions every time it runs.)

When the script is run with the option `--attributes`, it instead adds the availability incantation as an actual availability attribute:

```swift
/*System 0.0.2*/@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
public func greeting() -> String {
  "Hello"
}
```

This can be used by package maintainers to generate the ABI stable variant of swift-system that ships in Apple's OS distributions.

The script recognizes all three forms of these annotations and updates them on every run, so we can run the script to enable/disable attributes as needed.

### DIFF
--- a/Sources/System/Errno.swift
+++ b/Sources/System/Errno.swift
@@ -10,7 +10,7 @@
 /// An error number used by system calls to communicate what kind of error
 /// occurred.
 @frozen
-// @available(macOS 10.16, iOS 14.0, watchOS 7.0, tvOS 14.0, *)
+/*System 0.0.1, @available(macOS 11.0, iOS 14.0, watchOS 7.0, tvOS 14.0, *)*/
 public struct Errno: RawRepresentable, Error, Hashable, Codable {
   /// The raw C error number.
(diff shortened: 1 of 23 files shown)


lorentey:
@swift-ci test

milseman (approved):
LGTM!

I mean it's kinda crazy you have to do this, but keep doing it!

lorentey:
@swift-ci test

PR #80, opened 2022-04-08
Mitigate misuse of Swift's pointer conversion feature

glessard:
This is a continuation of the work done in https://github.com/apple/swift/pull/42002, addressing the exact same issues.

The family of `String` (and `FilePath`) initializers that convert from C strings (null-terminated byte buffers) can be called with Swift arrays, which are converted to `UnsafePointer` arguments for C interoperability. However, when the array passed in to them violates the C string precondition of containing a zero byte, this can result in a buffer overflow.

This PR overloads every such initializer with a version for `[CodeUnit]` and `inout CodeUnit`, enforcing the null-terminated precondition. An overload for String is also added. The String overload may appear strictly useless, but it behaves differently than a direct copy when the source string contains an embedded null.

Addresses rdar://91436410.

### DIFF
--- a/Sources/System/FilePath/FilePathString.swift
+++ b/Sources/System/FilePath/FilePathString.swift
@@ -20,6 +20,49 @@ extension FilePath {
     self.init(_platformString: platformString)
   }
 
+  /// Creates a file path by copying bytes from a null-terminated platform
+  /// string.
+  ///
+  /// - Note It is a precondition that `platformString` must be null-terminated.
+  /// The absence of a null byte will trigger a runtime error.
+  ///
+  /// - Parameter platformString: A null-terminated platform string.
+  @inlinable
+  @_alwaysEmitIntoClient
+  public init(platformString: [CInterop.PlatformChar]) {
+    guard let _ = platformString.firstIndex(of: 0) else {
+      fatalError(
+        "input of FilePath.init(platformString:) must be null-terminated"
+      )
+    }
+    self = platformString.withUnsafeBufferPointer {
+      FilePath(platformString: $0.baseAddress!)
+    }
+  }
+
+  @inlinable
+  @_alwaysEmitIntoClient
+  @available(*, deprecated, message: "Use FilePath(_: String) to create a path from a String.")
+  public init(platformString: inout CInterop.PlatformChar) {
+    guard platformString == 0 else {
+      fatalError(
+        "input of FilePath.init(platformString:) must be null-terminated"
+      )
+    }
+    self = FilePath()
+  }
+
+  @inlinable
+  @_alwaysEmitIntoClient
+  @available(*, deprecated, message: "Use FilePath(_: String) to create a path from a String.")
+  public init(platformString: String) {
+    if let nullLoc = platformString.firstIndex(of: "\0") {
+      self = FilePath(String(platformString[..<nullLoc]))
+    } else {
+      self = FilePath(platformString)
+    }
+  }
+
   /// Calls the given closure with a pointer to the contents of the file path,
   /// represented as a null-terminated platform string.
   ///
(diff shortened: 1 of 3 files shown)


glessard:
There are no new tests yet.

Sources/System/FilePath/FilePathString.swift
```diff
@@ -20,6 +20,30 @@ extension FilePath {
     self.init(_platformString: platformString)
   }
 
+  public init(platformString: [CInterop.PlatformChar]) {
```
milseman, line 23:
```suggestion
  public init<C: Collection>(platformString: C) where C.Element == CInterop.PlatformChar {
```

Sources/System/FilePath/FilePathString.swift
```diff
@@ -20,6 +20,30 @@ extension FilePath {
...
+    guard let _ = platformString.firstIndex(of: 0) else {
+      preconditionFailure(
+        "input of FilePath.init(platformString:) must be null-terminated"
```
milseman, line 26: If this is meant to take a collection of the contents, should we fail if there is an unexpected interior null?

Sources/System/FilePath/FilePathString.swift
```diff
@@ -20,6 +20,30 @@ extension FilePath {
...
+      )
+    }
+    self = platformString.withUnsafeBufferPointer {
+      FilePath(_platformString: $0.baseAddress!)
+    }
+  }
+
+  public init(platformString: inout CInterop.PlatformChar) {
```
milseman, line 34: What role is this init meant to serve?

Sources/System/FilePath/FilePathString.swift
```diff
@@ -96,6 +148,34 @@ extension FilePath.Root {
...
+  public init?(platformString: inout CInterop.PlatformChar) {
+    guard platformString == 0 else { return nil }
+    let root = withUnsafePointer(to: &platformString) {
+      FilePath.Root(_platformString: $0)
+    }
+    guard let root = root else { return nil }
+    self = root
+  }
+
+  public init?(platformString: String) {
```
milseman, line 171: What calls this?

Sources/System/FilePath/FilePathString.swift
glessard: My thought is that this should act the same as if it came as a pointer from C, and stop at the first null -- no matter where it is. It is _possible_ that there is code out there that relies on that behaviour.

Sources/System/FilePath/FilePathString.swift
glessard: That's another valid pointer conversion; ideally we'd mark this as deprecated. We can also not bother with it.

glessard: The ideal outcome is to restrict pointer conversions to C interop only, and add an attribute to opt in to pointer conversions for shims that call out to C functions. Obviously that's not happening this year.

Sources/System/FilePath/FilePathString.swift
glessard: In the stdlib, I marked the equivalent as deprecated with a suggestion to use `String.init(_: UnicodeScalar)` instead.

Sources/System/FilePath/FilePathString.swift
glessard: I'm not sure we want to go generic unless we ultimately want to keep that overload in place. Ideally this would be `@_alwaysEmitIntoClient` so that we can remove it once the language has evolved to eliminate the issue we're mitigating against. That's an interesting point of discussion…

Sources/System/FilePath/FilePathString.swift
milseman: What's the initializer used for? We'd at the very least want to put null-terminated in the argument label.

glessard: It's a mitigation for an array-to-pointer conversion that leads to reading out of bounds if the array does not contain a null. It has to be an exact overload for `init(platformString: UnsafePointer<CInterop.PlatformChar>)` in order to do its job.

Sources/System/FilePath/FilePathString.swift
milseman: Can you write a doc comment that carefully explains the null is expected here? Can you try AEIC? Would a deprecation annotation be possible?

Sources/System/FilePath/FilePathString.swift
milseman: Can we do deprecation too?

Sources/System/FilePath/FilePathString.swift
```diff
@@ -60,6 +84,34 @@ extension FilePath.Component {
     self.init(_platformString: platformString)
   }
 
+  public init?(platformString: [CInterop.PlatformChar]) {
+    if let _ = platformString.firstIndex(of: 0),
```
milseman, line 88: `guard`?

milseman: Also, 

> Components are never empty and never contain the directory separator.

So can we just return `nil`?

Sources/System/FilePath/FilePathString.swift
```diff
@@ -60,6 +84,34 @@ extension FilePath.Component {
...
+       }) {
+      self = component
+      return
+    }
+    return nil
+  }
+
+  public init?(platformString: inout CInterop.PlatformChar) {
+    guard platformString == 0 else { return nil }
+    let component = withUnsafePointer(to: &platformString) {
```
milseman, line 100: > Components are never empty and never contain the directory separator.

So can we just return `nil`?

milseman (approved):
Put doc comments and some cleanup, otherwise LGTM

Sources/System/FilePath/FilePathString.swift
glessard: I think we can. I did want your opinion on that.

Sources/System/FilePath/FilePathString.swift
glessard: For the FilePath ones I'm not sure what to put in the deprecation messages. It's pretty easy for String...

Sources/System/FilePath/FilePathString.swift
```diff
@@ -20,6 +20,30 @@ extension FilePath {
...
+    }
+  }
+
+  public init(platformString: inout CInterop.PlatformChar) {
+    guard platformString == 0 else {
+      preconditionFailure(
+        "input of FilePath.init(platformString:) must be null-terminated"
+      )
+    }
+    self = withUnsafePointer(to: &platformString, FilePath.init(_platformString:))
```
glessard, line 40: This just forces FilePath.init to build with an empty string. Is there a constant for an empty path?

glessard: That's `FilePath()`, I guess.

glessard:
@swift-ci please test

glessard:
@swift-ci please test

glessard:
@swift-ci please test

glessard:
@swift-ci please test

glessard:
@swift-ci please test

lorentey (approved):
Shame we need to add all these overloads -- can't wait to have a language-level solution.

glessard:
@swift-ci please test

PR #81, opened 2022-04-20
Update Linux tests manifest

dduan:
Run `swift package --generate-linuxmain` results in this patch.


dduan:
Side note: this could've been caught pretty easily with CI (Github Actions).

PR #82, opened 2022-05-04
FileDescriptor: Add resize(to newSize:) and fileSize() APIs

simonjbeaumont:
The `FileDescriptor` API currently doesn't expose anything that resembles `ftruncate(2)` for truncating or extending the file associated with a file descriptor.

This PR adds a new API `FileDescriptor.resize(to newSize:)` which wraps `ftruncate`, available on non-Windows platforms.

### DIFF
--- a/Sources/System/FileHelpers.swift
+++ b/Sources/System/FileHelpers.swift
@@ -122,4 +122,18 @@ extension FileDescriptor {
       return .success(buffer.count)
     }
   }
+
+  /// Return the current size of the file.
+  ///
+  /// - Returns: The current size of the file, in bytes.
+  @_alwaysEmitIntoClient
+  @discardableResult
+  public func fileSize(
+    retryOnInterrupt: Bool = true
+  ) throws -> Int64 {
+    let current = try seek(offset: 0, from: .current)
+    let size = try seek(offset: 0, from: .end)
+    try seek(offset: current, from: .start)
+    return size
+  }
 }
--- a/Sources/System/FileHelpers.swift
+++ b/Sources/System/FileHelpers.swift
@@ -46,9 +46,6 @@ extension FileDescriptor {
   /// increments that position by the number of bytes written.
   /// See also ``seek(offset:from:)``.
   ///
-  /// This method either writes the entire contents of `sequence`,
-  /// or throws an error if only part of the content was written.
-  ///
   /// If `sequence` doesn't implement
   /// the <doc://com.apple.documentation/documentation/swift/sequence/3128824-withcontiguousstorageifavailable> method,
   /// temporary space will be allocated as needed.
--- a/Sources/System/FileHelpers.swift
+++ b/Sources/System/FileHelpers.swift
@@ -122,18 +122,4 @@ extension FileDescriptor {
       return .success(buffer.count)
     }
   }
-
-  /// Return the current size of the file.
-  ///
-  /// - Returns: The current size of the file, in bytes.
-  @_alwaysEmitIntoClient
-  @discardableResult
-  public func fileSize(
-    retryOnInterrupt: Bool = true
-  ) throws -> Int64 {
-    let current = try seek(offset: 0, from: .current)
-    let size = try seek(offset: 0, from: .end)
-    try seek(offset: current, from: .start)
-    return size
-  }
 }
--- a/Sources/System/FileOperations.swift
+++ b/Sources/System/FileOperations.swift
@@ -370,6 +370,47 @@ extension FileDescriptor {
   public func dup2() throws -> FileDescriptor {
     fatalError("Not implemented")
   }
+
+  /// Truncate or extend the file referenced by this file descriptor.
+  ///
+  /// - Parameters:
+  ///   - newSize: The length in bytes to resize the file to.
+  ///   - retryOnInterrupt: Whether to retry the write operation
+  ///      if it throws ``Errno/interrupted``. The default is `true`.
+  ///      Pass `false` to try only once and throw an error upon interruption.
+  ///
+  /// The file referenced by this file descriptor will by truncated (or extended) to `newSize`.
+  ///
+  /// If the current size of the file exceeds `newSize`, any extra data is discarded. If the current
+  /// size of the file is smaller than `newSize`, the file is extended and filled with zeros to the
+  /// provided size.
+  ///
+  /// This function requires that the file has been opened for writing.
+  ///
+  /// - Note: This function does not modify the current offset for any open file descriptors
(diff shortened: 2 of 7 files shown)


Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -105,4 +105,11 @@ internal func pwrite(
   return Int(nNumberOfBytesWritten)
 }
 
+@inline(__always)
+internal func ftruncate(
+  _ fd: Int32, _ length: off_t
+) -> Int32 {
+  _chsize(fd, numericCast(length))
+}
```
simonjbeaumont, line 113: Full disclosure: I have not tested the Windows path. AFAICT there isn't a Windows CI pipeline?

simonjbeaumont:
@swift-ci please test

simonjbeaumont:
@swift-ci please test

milseman:
@lorentey can you take an initial look?

Sources/System/FileHelpers.swift
```diff
@@ -122,4 +122,18 @@ extension FileDescriptor {
       return .success(buffer.count)
     }
   }
+
+  /// Return the current size of the file.
+  ///
+  /// - Returns: The current size of the file, in bytes.
+  @_alwaysEmitIntoClient
+  @discardableResult
```
simonjbeaumont, line 130: Looking at this again, this probably shouldn't be discardable.

Sources/System/FileHelpers.swift
```diff
@@ -122,4 +122,18 @@ extension FileDescriptor {
...
+  public func fileSize(
+    retryOnInterrupt: Bool = true
+  ) throws -> Int64 {
+    let current = try seek(offset: 0, from: .current)
+    let size = try seek(offset: 0, from: .end)
+    try seek(offset: current, from: .start)
```
lorentey, line 136: This is temporarily mutating the file position during what looks like a read-only operation; I feel the potential pitfalls of this (alongside the fact that many file descriptors do not have a file size) probably makes this entry point less suitable for this package.

Sources/System/Internals/WindowsSyscallAdapters.swift
lorentey: I believe it's best to surround the addition of `resize` with `#if !os(Windows)`.

Generally, I don't think it's a good idea to try to shoehorn Windows support into POSIX code paths here, or anywhere else in this code base. Swift System is not intended to provide a portable API layer, and the Windows SDK is different enough that it deserves its own, standalone, implementations.

lorentey (approved):
Big thumbs up on `resize`! I'm less convinced we need `fileSize`.

Let's keep the Windows implementation out for now.

Sources/System/FileHelpers.swift
simonjbeaumont: That's reasonable. I added the following fixup to address this:

* d90bf85 `fixup: Move FileDescriptor.fileSize() to an internal extension in SystemTests`

Sources/System/Internals/WindowsSyscallAdapters.swift
simonjbeaumont: OK, apologies for that, I was just following the pattern I saw with the existing `WindowsSyscallAdapters`. I added some fixups which move this API to only non-Windows platforms:

* 02daf4b `fixup: Move FileDescriptor.resize to extension with #if !os(Windows)`
* 7fab65f `fixup: Remove Windows syscall adaptor for ftruncate`

Sources/System/FileHelpers.swift
simonjbeaumont: Addressed as part of:

* d90bf85 `fixup: Move FileDescriptor.fileSize() to an internal extension in SystemTests`

simonjbeaumont (commented):
Thanks @lorentey for the review!

> Big thumbs up on `resize`! I'm less convinced we need `fileSize`.

No problem, I moved `fileSize` to be an implementation detail of the tests for `resize`.

> Let's keep the Windows implementation out for now.

Done.

Sources/System/FileOperations.swift
```diff
@@ -398,3 +398,51 @@ extension FileDescriptor {
...
+  /// size of the file is smaller than `newSize`, the file is extended and filled with zeros to the
+  /// provided size.
+  ///
+  /// This function requires that the file has been opened for writing.
+  ///
+  /// - Note: This function does not modify the current offset for any open file descriptors
+  /// associated with the file.
+  ///
+  /// The corresponding C function is `ftruncate`.
+  /*System 1.2.0, @available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)*/
```
milseman, line 425: @lorentey what's the right availability comment here?

simonjbeaumont: Yeah, I took a punt here based on what I saw with previous evolution and given there already exists a public `1.1.0` tag. LMK if it needs changing.

simonjbeaumont:
@swift-ci please test

Sources/System/FileOperations.swift
simonjbeaumont: @lorentey Could you advise on what this magic comment string should be? I took a look at `./Utilities/expand-availability.py` and now I think maybe it should just be the following:

```swift
/*System 1.2.0*/
```

Happy to update this to reflect your release workflow if you can point me in the right direction here.

simonjbeaumont: @lorentey LMK what this should be and I can update this PR.

lorentey: Ah, apologies for the delay -- `/*System 1.2.0*/` is going to be the right one, it just needs to be defined in `/Utilities/expand-availability.py`. (Expanded to 9999 versions.)

We can define the version directly within this PR, or we can just land it with an undefined version, and let maintainers deal with it later. (These comments do not matter within this package variant of System.)

lorentey: (To clarify, the comment we have here is just fine -- we don't need to remove the `@available` incantation.)

simonjbeaumont: Added in:

* ca6e888 `Utilities: Add version 1.2.0 to expand-availability.py tool`

simonjbeaumont:
@swift-ci please test

Sources/System/FileOperations.swift
lorentey: How did I miss it! 👍

simonjbeaumont:
@swift-ci please test

PR #86, opened 2022-05-19
FileOperations: Fix _open precondition

dduan:
I _think_ this precondition is filpped. Without this patch, the following code would fail the precondition:

```swift
try FileDescriptor.open(somePath, .writeOnly, options: [.create])
```

While it should translate to `O_WRONLY | O_CREAT`, a very common flag.

(I also think this precondition's existence is debatable. But I understand it guards against some surprises for those unfamiliar with this corner of POSIX)

### DIFF
--- a/Sources/System/FileOperations.swift
+++ b/Sources/System/FileOperations.swift
@@ -77,7 +77,7 @@ extension FileDescriptor {
       if let permissions = permissions {
         return system_open(path, oFlag, permissions.rawValue)
       }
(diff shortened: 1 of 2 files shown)


dduan:
@swift-ci test

dduan:
AH, the sentence `Create must be given permissions` didn't make sense 6 hours ago. It did now that I got a little sleep 🥲. Sorry about that!

PR #89, opened 2022-06-06
FileOperations: Add Windows implementation for FileDescriptor.resize(to:)

simonjbeaumont:
#82 added `FileDescriptor.resize(to:)`. The original PR included adding `ftruncate` to `WindowsSycallAdapters.swift` but it was [removed during review](https://github.com/apple/swift-system/pull/82#discussion_r866323055), and so there is currently no implementation for `resize(to:)` on Windows.

The justification was that we didn't want to shoehorn Windows support into the POSIX code paths and that the Windows SDK was different enough to justify its own platform-specific implementation. However, looking at `chsize` (renamed `_chsize`) it looks like it's supposed to be the equivalent to the POSIX-like functions in UCRT.

In order to move discussion forward on an appropriate Windows implementation I have reinstated the syscall adapter as a strawman proposal.

PR #90, opened 2022-06-09
fixup: Remove os(Windows) guards for resize internals

simonjbeaumont:
#89 attempted to add `FileDescriptor.resize(to:)` for Windows too, but there was still an `#if !os(Windows)` guard around an implementation detail so the Windows build failed.

This PR removes the guards around the internal implementation and the tests.


milseman:
@swift-ci please test

compnerd:
@swift-ci please test

PR #91, opened 2022-06-10
Tests: ignore XCTest manifest on newer Swift

compnerd:
This is motivated by the desire to build the test suite on Windows.
Newer versions of SPM have test discovery which obviates the test
manifest, which incidentally repairs the build on Windows while adopting
newer functionality and reducing management complexity.

### DIFF
--- a/Tests/SystemTests/XCTestManifests.swift
+++ b/Tests/SystemTests/XCTestManifests.swift
@@ -1,4 +1,4 @@
-#if !canImport(ObjectiveC)
+#if !canImport(ObjectiveC) && swift(<5.5)
 import XCTest
 
 extension ErrnoTest {


compnerd:
@swift-ci please test

compnerd:
Yes, I was conflicted about that.  It seems that it would be better to remove the file - it’s unlikely it will be maintained, test discovery is now default even, so why keep it?  But perhaps there are older environments that need to be supported?

milseman:
We'll likely be bumping the tools version soon. I want to get a `FileDescriptor.readAll() -> some RAC<UInt8>`, etc.

compnerd:
That is fair; what should we do in the meantime?  I would like to repair the test suite for Windows.

Lukasa:
@milseman Is your plan to bump the tools version all the way to 5.7?

lorentey:
@swift-ci test

PR #95, opened 2022-06-11
Tests: explicitly convert the character type

compnerd:
The platform character type is different across Unix and Windows.
Explicitly convert the value to the expected type.

### DIFF
--- a/Tests/SystemTests/SystemCharTest.swift
+++ b/Tests/SystemTests/SystemCharTest.swift
@@ -25,7 +25,7 @@ final class SystemCharTest: XCTestCase {
 
     // non printable
     for value in 0..<(UInt8(ascii: " ")) {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/SystemTests/SystemCharTest.swift
```diff
@@ -25,7 +25,7 @@ final class SystemCharTest: XCTestCase {
 
     // non printable
     for value in 0..<(UInt8(ascii: " ")) {
-      XCTAssertFalse(SystemChar(codeUnit: value).isLetter)
```
lorentey, line 28: 🤦‍♂️ What exactly are we trying to achieve with these APIs, @milseman?

PR #96, opened 2022-06-12
Tests: remove error codes that are unavailable on windows

compnerd:
Not all the error codes are available on Windows, this matches
SystemErrno cases to remove the cases that have no definition on
Windows.


compnerd:
@swift-ci please test

lorentey (approved):
I am noting sincere doubt that these tests are achieving any function other than making it more difficult to maintain this package.

PR #97, opened 2022-06-12
Tests: explicitly construct `FilePath`

compnerd:
The literal value needs to be converted on different platform.
Explicitly construct the type rather than trying to change the literal
to the platform character type.

### DIFF
--- a/Tests/SystemTests/FileOperationsTest.swift
+++ b/Tests/SystemTests/FileOperationsTest.swift
@@ -163,7 +163,7 @@ final class FileOperationsTest: XCTestCase {
 
 #if !os(Windows)
   func testResizeFile() throws {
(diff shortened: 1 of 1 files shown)


Tests/SystemTests/FileOperationsTest.swift
```diff
@@ -163,7 +163,7 @@ final class FileOperationsTest: XCTestCase {
 
 #if !os(Windows)
   func testResizeFile() throws {
-    let fd = try FileDescriptor.open("/tmp/\(UUID().uuidString).txt", .readWrite, options: [.create, .truncate], permissions: .ownerReadWrite)
+    let fd = try FileDescriptor.open(FilePath("/tmp/\(UUID().uuidString).txt"), .readWrite, options: [.create, .truncate], permissions: .ownerReadWrite)
```
lorentey, line 166: Hm. This feels like it indicates an API bug.

milseman: The string-taking version should either be disfavored or updated to construct a file path.

PR #98, opened 2022-06-12
Tests: explicitly reference `PlatformChar` rather than `CChar`

compnerd:
There is no guarantee that the `PlatformChar` is `CChar`.  On Windows,
this needs to be UTF16 rather than UTF8.  Explicitly use the
`PlatformChar` rather than `CChar`.

### DIFF
--- a/Tests/SystemTests/FilePathTests/FilePathTest.swift
+++ b/Tests/SystemTests/FilePathTests/FilePathTest.swift
@@ -22,7 +22,7 @@ func filePathFromUnterminatedBytes<S: Sequence>(_ bytes: S) -> FilePath where S.
   array += [0]
 
   return array.withUnsafeBufferPointer {
-    $0.withMemoryRebound(to: CChar.self) {
+    $0.withMemoryRebound(to: CInterop.PlatformChar.self) {
       FilePath(platformString: $0.baseAddress!)
     }
   }


Tests/SystemTests/FilePathTests/FilePathTest.swift
```diff
@@ -22,7 +22,7 @@ func filePathFromUnterminatedBytes<S: Sequence>(_ bytes: S) -> FilePath where S.
   array += [0]
 
   return array.withUnsafeBufferPointer {
-    $0.withMemoryRebound(to: CChar.self) {
```
lorentey, line 25: This probably isn't right -- I expect we'll need an alternate function that takes wchar_t input.

Preferably though we should just mark these tests as not supporting Windows, and eventually set up dedicated tests for that platform.

compnerd: Am I understanding what you are suggesting correctly here?
```swift
func filePathFromUnterminatedBytes<S: Sequence>(_ bytes: S) -> FilePath where S.Element == UInt8
func filePathFromUnterminatedBytes<S: Sequence>(_ bytes: S) -> FilePath where S.Element == UInt16
```
This seems to make sense to me (at least more than the alternative interpretation where the `FilePath` initializer explicitly disambiguates).

PR #102, opened 2022-06-12
Tests: adjust the Windows error codes

compnerd:
The Windows error codes are not re-exported, explicitly import the
WinSDK module to gain access to the error codes.  Add Windows specific
checks to ensure that the named forms match the underlying
implementation.

### DIFF
--- a/Tests/SystemTests/ErrnoTest.swift
+++ b/Tests/SystemTests/ErrnoTest.swift
@@ -15,6 +15,10 @@ import SystemPackage
 import System
 #endif
 
+#if os(Windows)
+import WinSDK
+#endif
+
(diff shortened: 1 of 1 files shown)


lorentey (approved):
+1 to make this test pass on Windows. However, why do we have `Errno` there in the first place? Does it really make sense to ELOOP and WSAEPFNOSUPPORT both handled by the same error type?

lorentey:
@swift-ci test

compnerd:
Well, I can understand having different families of error codes, but that is the status quo on all the platforms - all the error codes are in a single domain.  Splitting them up based on usage seems interesting but also problematic for things like sockets on unixy platforms.

lorentey:
Note: I'm not saying that `Errno` definitely wasn't the right choice on Windows -- but I do strongly suspect that is probably the case.

This should not be a thing:

```
    // TODO(compnerd) map windows error to errno
```

PR #103, opened 2022-06-12
Tests: change the conversion for string to array

compnerd:
Windows uses UTF16 for the platform character, not UTF8, and does not
map to CChar.  Add a Windows path for the conversion.


lorentey (approved):

lorentey:
@swift-ci test

PR #105, opened 2022-06-21
Remove Windows warnings

dduan:
These calls are missing their Windows counter parts, resulting warnings stating `X is deprecated: The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: Y. See online help for details`.

Add them `Y`s like the rest.


milseman:
@compnerd can you take a look?

compnerd (approved):
The mapping looks correct to me (and is what Microsoft recommends).

dduan:
Thanks!

PR #109, opened 2022-09-10
Quote cmake include path

krzyzanowskim:
quote include path that helps if path contains eg. a space

### DIFF
--- a/cmake/modules/SwiftSystemConfig.cmake.in
+++ b/cmake/modules/SwiftSystemConfig.cmake.in
@@ -8,5 +8,5 @@ See https://swift.org/LICENSE.txt for license information
 #]]
 
 if(NOT TARGET SystemPackage)
-  include(@SWIFT_SYSTEM_EXPORTS_FILE@)
+  include("@SWIFT_SYSTEM_EXPORTS_FILE@")
 endif()


milseman:
@compnerd can you take a look?

compnerd (approved):
I think that this could use a better commit message - it isn't immediately clear what it means.  Normally the include path would refer to the `CMAKE_MODULE_PATH` which shouldn't require quoting if handled properly IIRC.  This is in the generated config file and the quoting does matter here.

krzyzanowskim:
> refer to the CMAKE_MODULE_PATH which shouldn't require quoting if handled properly IIRC

long story short: swift won't build if checkout to path with space

When building swift-project using [build-script](https://github.com/apple/swift/blob/main/utils/build-script) if fails randomly here and there, and this is one place where it fail to include cmake macros from a path with spaces.

compnerd:
@krzyzanowskim I think that you misunderstood me, I agree that this site should be quoted.  I think that the "quote cmake include path" is a bit confusing is all.  This is an include, but the quoting is applied to a parameter.  The "CMake include path" is normally the `CMAKE_MODULE_PATH`, which is not being touched here and is confusing because you expect the diff to be different.

compnerd:
@swift-ci please test

compnerd:
I don't think that the commit message is worth keeping this open.  Going to merge it.

PR #112, opened 2022-10-25
Add Sendable conformances

ffried:
This adds `Sendable` conformances to all public types.


Sources/System/FileDescriptor.swift
```diff
@@ -473,3 +473,10 @@ extension FileDescriptor.OpenOptions
   /// A textual representation of the open options, suitable for debugging.
   public var debugDescription: String { self.description }
 }
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+extension FileDescriptor: Sendable {}
```
Lukasa, line 478: I'm not convinced `FileDescriptor` should be `Sendable`: it has a bunch of global state that can absolutely misbehave if you concurrently use it from multiple threads.

ffried: Oh, good catch!
Should I annotate the conformance as `@available(*, unavailable)` to make this explicit?

ffried: I think it would make sense. The correct way would be to pass the `FilePath` across threads and open a `FileDescriptor` on each thread that needs one.

I've added the annotation.

ffried: @milseman I completely agree with your reasoning here. What I'm still unsure about is, how this will behave in future Swift versions. I'm not too familiar with the plans there, but AFAIK Swift 6 will warn when passing around non-`Sendable` objects across async contexts. Is there a difference between _not_ conforming `FileDescriptor` to `Sendable` and explicitly making it "non-Sendable"?
The effect would still be the same (having to pass around `rawValue`), or am I missing something here?

Note that I don't have an issue with removing the conformance altogether. I'm wondering whether this is (or will be) actually different from explicitly marking it as unavailable.

lorentey: I agree too! `FileDescriptor` is exactly the same thing as the corresponding integer value, just wrapped in nicer clothing. We can freely pass integers across async contexts, so we should be able to do the same with `FileDescriptor` instances.

Passing file descriptors across actor boundaries may or may not be a good idea in general -- as usual, it depends on the actual use case. In any case, we certainly do not want to force people to do contortions like extracting the underlying integer to be able to do this.

lorentey: (To put it another way, if we want to prevent people from using raw syscalls like `read` and `write` in certain async contexts, then we should actually figure out how to do that, rather than making the Swift wrappers more frustrating to use.)

ffried: I think we have three options here (and in this discussion also at least partial votes for all three of them 😄):

1. Make `FileDescriptor` conform to `Sendable`
2. Don't add any `Sendable` conformance to `FileDescriptor`
3. Add it but mark it as `@available(*, unavailable)`

I'm a bit torn between 1 and 3.
Given that `FileDescriptor`s *can* be used in a safe way across threads, making them explicitly **not** `Sendable` could prevent some clean code - unless those "safe ways" are represented by a different, explicitly `Sendable` type.
However, since I think this library also aims to prevent "common errors" with the underlying syscalls, declaring `FileDescriptor` as `Sendable` could send the wrong signal. I'm just not sure if this is still in the scope of the safety features this package aims to provide.

Option 2 isn't really a good solution IMHO, since it only delays the decision. I'd be fine, however, with going with option 2 in this PR and opening separate PR for the discussion about `FileDescriptor`.

For now I've re-added the conformance and added a comment above it.

karwa: > FileDescriptor is exactly the same thing as the corresponding integer value, just wrapped in nicer clothing. We can freely pass integers across async contexts, so we should be able to do the same with FileDescriptor instances.

Wouldn't this reasoning suggest that unsafe pointers should be Sendable again?

Lukasa: I'm with @karwa, and I'll escalate somewhat and point out to @lorentey that he and I discussed whether `ManagedAtomic` should be `Sendable`. @lorentey made what I considered a fairly compelling argument that it shouldn't be, noting that it is inappropriate for the compiler to assert that a type holding a `ManagedAtomic` is `Sendable`.

IMO users who wish to use `FileDescriptor` in a way that is `Sendable` can wrap it in a type that enforces the invariant and mark it `@unchecked Sendable`. Otherwise, users rightly should be discouraged from sharing a `FileDescriptor` across threads and calling `read` on it.

ffried: @Lukasa Does this mean that the future locking features that @milseman mentioned should be implemented using a new type (wrapping `FileDescriptor`) that is `@unchecked Sendable`?

Lukasa: I think if we have a type using file locks then yes, I'd suggest that being a different type that we can construct from and transform into the unsendable type.

ffried: Alright, I've marked the `Sendable` conformance as `@available(*, unavailable)` (again 😅).
Also adjusted the comment above it to explain the reasoning.

ffried: @milseman Thank you for this in-depth explanation! 

I'm not sure if I can provide any useful input to this decision. Even - or especially - with your write-up, I still see both options as reasonable. Either option will need a proper documentation explaining the decision and its consequences.

> I'd like to debate/stew over this a little more and figure out how it could fit in with System's philosophy, especially as we see how file locks, descriptors shared across fork\exec, etc. work out.

Would it make sense to factor the conformance of `FileDescriptor` out into a separate PR for this?
Or will this just create confusion in case a release is finalized w/o the conformance decision for `FileDescriptor`?

MaxDesiatov: Sorry for interjecting, but there are other projects that would benefit from `FilePath: Sendable` conformance, so I personally hope at least that part could land soon-ish if there are no objections to it.

milseman: I'm fine splitting FileDescriptor's decision off or keeping this open, your call.

Another useful point, would socket descriptors be sendable @Lukasa? Our prototype had a different type for sockets with explicit conversions between file and socket descriptors. Similarly, read-ends of pipes, etc. We might end up with a more fleshed out descriptor story with multiple types.

Lukasa: That's a good question with a tricky answer. By the same logic that file descriptors aren't, socket descriptors aren't (calling `read` from multiple threads necessarily has unexpected effects). Ideally we'd be able to move it across threads but not to share the reference. This would encourage users to `dup` the FD, which has substantially clearer semantics.

ffried: I've split off the other conformances in #115. This PR now only adds the FileDescriptor conformance (on top of #115).

milseman:
I'd like to reengage with this topic. I'm still (even more) on the side of `FileDescriptor` being `Sendable`. @lorentey what would be the ABI impact of adding that conformance?

ffried:
@milseman @lorentey Should I change this PR to _add_ the conformance (w/o marking it as unavailable)?

milseman:
Yes to adding it, deferring to @lorentey regarding availability.

ffried:
In accordance with the change in 87e68fc, I've updated the conformance to be on the main type. This also resolves the conflicts that existed on this PR's branch.

Depending on @lorentey's decision regarding availability, I'll move it to an extension again (for marking it unavailable).

lorentey:
Update: As a `@frozen` struct wrapping a single integer, the shipping `FileDescriptor` has already been implicitly inferred to be `Sendable`, as it should be.

Concurrent use of the same file descriptor does not lead to undefined behavior. And even if it did, it is not in Swift System's mandate to protect against that.

The job Swift System is merely to present the preexisting C APIs in a way that feels native to Swift. This codebase is not in the business of inventing nontrivial abstractions over the constructs provided by the operating system, or of building pretend obstacles against their use. One can pass integer values between concurrent execution contexts and directly call the `read`/`write`/`close`/etc. syscalls on them -- therefore, it must be possible to do the same on `FileDescriptor` instances and their operations.

ffried:
@lorentey Thanks for the update. In this case this PR should be ready as is, right? I've just rebased the branch on main to make sure there are no conflicts.

lorentey (approved):

lorentey:
@swift-ci test

PR #113, opened 2022-10-29
Support OpenBSD.

3405691582:
Add the usual `os` conditionals and exclude errno constants that aren't present on this platform.

Maybe availability annotations are a better way to deal with this, but that is likely something that requires more careful consideration.


milseman:
LGTM.

@rauhul I know you've been interested in non-Darwin BSD variants. Could you glance over this PR and let me know what you think?

@swift-ci please test

rauhul:
@milseman This looks all correct to me, I did a quick (non-exhaustive check against (OpenBSD's errno.h)[https://github.com/openbsd/src/blob/master/sys/sys/errno.h].

PR #116, opened 2022-12-08
Add Mach.Port

loffgren:
This is a new Swift move-only type interface to Mach ports.

### DIFF
--- a/Sources/System/MachPort.swift
+++ b/Sources/System/MachPort.swift
@@ -1,0 +1,318 @@
+/*
+ This source file is part of the Swift System open source project
+
+ Copyright (c) 2022 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+#if $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
+
+import Darwin.Mach
+
+protocol MachPortRight {}
+
+enum Mach {
+    @_moveOnly
+    struct Port<RightType:MachPortRight> {
+        internal var name:mach_port_name_t
+        internal var context:mach_port_context_t
+
+        /// Transfer ownership of an existing unmanaged Mach port right into a
+        /// Mach.Port by name.
+        ///
+        /// This initializer aborts if name is MACH_PORT_NULL.
+        ///
+        /// If the type of the right does not match the type T of Mach.Port<T>
+        /// being constructed, behavior is undefined.
+        ///
+        /// The underlying port right will be automatically deallocated at the
+        /// end of the Mach.Port instance's lifetime.
+        ///
+        /// This initializer makes a syscall to guard the right.
+        init(name:mach_port_name_t) {
+            assert(name != mach_port_name_t(MACH_PORT_NULL))
+            self.name = name
+
+            if (RightType.self == ReceiveRight.self) {
+                let secret = mach_port_context_t(arc4random())
+                let kr = mach_port_guard(mach_task_self_, name, secret, 0)
+                assert(kr == KERN_SUCCESS)
+                self.context = secret
+            }
+            else {
+                self.context = 0
+            }
+        }
+
+        /// Borrow access to the port name in a block that can perform
+        /// non-consuming operations.
+        ///
+        /// Take care when using this function; many operations consume rights,
+        /// and send-once rights are easily consumed.
+        ///
+        /// If the right is consumed, behavior is undefined.
+        ///
+        /// The body block may optionally return something, which will then be
+        /// returned to the caller of withBorrowedName.
+        func withBorrowedName<ReturnType>(body:(mach_port_name_t) -> ReturnType) -> ReturnType {
+            return body(name)
+        }
+
+        deinit {
+            if name != 0xFFFFFFFF /* MACH_PORT_DEAD */ {
+                if RightType.self == ReceiveRight.self {
+                    // recv rights must be mod ref'ed instead of deallocated
+                    let kr = mach_port_unguard(mach_task_self_, name, context)
+                    assert(kr == KERN_SUCCESS)
+
+                    let kr2 = mach_port_mod_refs(mach_task_self_, name, MACH_PORT_RIGHT_RECEIVE, -1)
+                    assert(kr2 == KERN_SUCCESS)
+                } else {
+                    mach_port_deallocate(mach_task_self_, name)
+                }
+            }
+        }
+    }
+
+    /// Possible errors that can be thrown by Mach.Port operations.
+    enum PortRightError : Error {
+        /// Returned when an operation cannot be completed, because the Mach
+        /// port right has become a dead name. This is caused by deallocation of the
+        /// receive right on the other end.
+        case deadName
+    }
+
+    /// The MachPortRight type used to manage a receive right.
+    struct ReceiveRight : MachPortRight {}
+
+    /// The MachPortRight type used to manage a send right.
+    struct SendRight : MachPortRight {}
+
+    /// The MachPortRight type used to manage a send-once right.
+    ///
+    /// Send-once rights are the most restrictive type of Mach port rights.
+    /// They cannot create other rights, and are consumed upon use.
+    ///
+    /// Upon destruction a send-once notification will be sent to the
+    /// receiving end.
+    struct SendOnceRight : MachPortRight {}
+
+    /// Create a connected pair of rights, one receive, and one send.
+    ///
+    /// This function will abort if the rights could not be created.
+    /// Callers may assert that valid rights are always returned.
+    static func allocatePortRightPair() -> (Mach.Port<Mach.ReceiveRight>, Mach.Port<Mach.SendRight>) {
+        var name = mach_port_name_t(MACH_PORT_NULL)
+        let secret = mach_port_context_t(arc4random())
+        withUnsafeMutablePointer(to: &name) { name in
+            var options = mach_port_options_t()
+            options.flags = UInt32(MPO_INSERT_SEND_RIGHT);
+            withUnsafeMutablePointer(to: &options) { options in
+                let kr = mach_port_construct(mach_task_self_, options, secret, name)
+                assert(kr == KERN_SUCCESS)
+            }
+        }
+        return (Mach.Port<Mach.ReceiveRight>(name: name, context: secret), Mach.Port<Mach.SendRight>(name: name))
+    }
+}
+
+extension Mach.Port where RightType == Mach.ReceiveRight {
+    /// Transfer ownership of an existing, unmanaged, but already guarded,
+    /// Mach port right into a Mach.Port by name.
+    ///
+    /// This initializer aborts if name is MACH_PORT_NULL.
+    ///
+    /// If the type of the right does not match the type T of Mach.Port<T>
+    /// being constructed, the behavior is undefined.
+    ///
+    /// The underlying port right will be automatically deallocated when
+    /// the Mach.Port object is destroyed.
+    init(name:mach_port_name_t, context:mach_port_context_t) {
+        self.name = name
+        self.context = context
+    }
+
+    /// Allocate a new Mach port with a receive right, creating a
+    /// Mach.Port<Mach.ReceiveRight> to manage it.
+    ///
+    /// This initializer will abort if the right could not be created.
+    /// Callers may assert that a valid right is always returned.
+    init() {
+        var storage:mach_port_name_t = 0
+        withUnsafeMutablePointer(to:&storage) { storage in
+            let kr = mach_port_allocate(mach_task_self_, MACH_PORT_RIGHT_RECEIVE, storage)
+            assert(kr == KERN_SUCCESS)
+        }
+
+        // name-only init will guard ReceiveRights
+        self.init(name:storage)
+    }
+
+
+    /// Transfer ownership of the underlying port right to the caller.
+    ///
+    /// Returns a tuple containing the Mach port name representing the right,
+    /// and the context value used to guard the right.
+    ///
+    /// This operation liberates the right from management by the Mach.Port,
+    /// and the underlying right will no longer be automatically deallocated.
+    ///
+    /// After this function completes, the Mach.Port is destroyed and no longer
+    /// usable.
+    __consuming func relinquish() -> (mach_port_name_t, mach_port_context_t) {
+        return (name, context)
+    }
+
+    /// Remove guard and transfer ownership of the underlying port right to
+    /// the caller.
+    ///
+    /// Returns the Mach port name representing the right.
+    ///
+    /// This operation liberates the right from management by the Mach.Port,
+    /// and the underlying right will no longer be automatically deallocated.
+    ///
+    /// After this function completes, the Mach.Port is destroyed and no longer
+    /// usable.
+    ///
+    /// This function makes a syscall to remove the guard from
+    /// Mach.ReceiveRights. Use relinquish() to avoid the syscall and extract
+    /// the context value along with the port name.
+    __consuming func unguardAndRelinquish() -> mach_port_name_t {
+        let kr = mach_port_unguard(mach_task_self_, name, context);
+        assert(kr == KERN_SUCCESS)
+        return name
+    }
+
+    /// Borrow access to the port name in a block that can perform
+    /// non-consuming operations.
+    ///
+    /// Take care when using this function; many operations consume rights.
+    ///
+    /// If the right is consumed, behavior is undefined.
+    ///
+    /// The body block may optionally return something, which will then be
+    /// returned to the caller of withBorrowedName.
+    func withBorrowedName<ReturnType>(body:(mach_port_name_t, mach_port_context_t) -> ReturnType) -> ReturnType {
+        body(name, context)
+    }
+
+    /// Create a send-once right for a given receive right.
+    ///
+    /// This does not affect the makeSendCount of the receive right.
+    ///
+    /// This function will abort if the right could not be created.
+    /// Callers may assert that a valid right is always returned.
+    func makeSendOnceRight() -> Mach.Port<Mach.SendOnceRight> {
+        // send once rights do not coalesce
+        var kr:kern_return_t = KERN_FAILURE
+        var newRight:mach_port_name_t = mach_port_name_t(MACH_PORT_NULL)
+        var newRightType:mach_port_type_t = MACH_PORT_TYPE_NONE
+
+        withUnsafeMutablePointer(to: &newRight) { newRight in
+            withUnsafeMutablePointer(to: &newRightType) { newRightType in
+                kr = mach_port_extract_right(mach_task_self_, name, mach_msg_type_name_t(MACH_MSG_TYPE_MAKE_SEND_ONCE), newRight, newRightType)
+
+            }
+        }
+
+        // The value of newRight is validated by the Mach.Port initializer
+        assert(kr == KERN_SUCCESS)
+        assert(newRightType == MACH_MSG_TYPE_MOVE_SEND_ONCE)
+
+        return Mach.Port<Mach.SendOnceRight>(name:newRight)
+    }
+
+    /// Create a send right for a given receive right.
+    ///
+    /// This increments the makeSendCount of the receive right.
+    ///
+    /// This function will abort if the right could not be created.
+    /// Callers may assert that a valid right is always returned.
+    func makeSendRight() -> Mach.Port<Mach.SendRight> {
+        let how = MACH_MSG_TYPE_MAKE_SEND
+
+        // send and recv rights are coalesced
+        let kr = mach_port_insert_right(mach_task_self_, name, name, mach_msg_type_name_t(how))
+        assert(kr == KERN_SUCCESS)
+
+        return Mach.Port<Mach.SendRight>(name:name)
+    }
+
+    /// Access the make-send count.
+    ///
+    /// Each get/set of this property makes a syscall.
+    var makeSendCount : mach_port_mscount_t {
+        get {
+            var status:mach_port_status = mach_port_status()
+            var size:mach_msg_type_number_t = mach_msg_type_number_t(MemoryLayout<mach_port_status>.size)
+            withUnsafeMutablePointer(to: &size) { size in
+                withUnsafeMutablePointer(to: &status) { status in
+                    let info = UnsafeMutableRawPointer(status).bindMemory(to: integer_t.self, capacity: 1)
+                    let kr = mach_port_get_attributes(mach_task_self_, name, MACH_PORT_RECEIVE_STATUS, info, size)
+                    assert(kr == KERN_SUCCESS);
+                }
+            }
+            return status.mps_mscount
+        }
+
+        set {
+            let kr = mach_port_set_mscount(mach_task_self_, name, newValue)
+            assert(kr == KERN_SUCCESS)
+        }
+    }
+}
+
+extension Mach.Port where RightType == Mach.SendRight {
+    /// Transfer ownership of the underlying port right to the caller.
+    ///
+    /// Returns the Mach port name representing the right.
+    ///
+    /// This operation liberates the right from management by the Mach.Port,
+    /// and the underlying right will no longer be automatically deallocated.
+    ///
+    /// After this function completes, the Mach.Port is destroyed and no longer
+    /// usable.
+    __consuming func relinquish() -> mach_port_name_t {
+        return name
+    }
+
+    /// Create another send right from a given send right.
+    ///
+    /// This does not affect the makeSendCount of the receive right.
+    ///
+    /// If the send right being copied has become a dead name, meaning the
+    /// receiving side has been deallocated, then copySendRight() will throw
+    /// a Mach.PortRightError.deadName error.
+    func copySendRight() throws -> Mach.Port<Mach.SendRight> {
+        let how = MACH_MSG_TYPE_COPY_SEND
+
+        // send rights are coalesced
+        let kr = mach_port_insert_right(mach_task_self_, name, name, mach_msg_type_name_t(how))
+        if kr == KERN_INVALID_CAPABILITY {
+            throw Mach.PortRightError.deadName
+        }
+        assert(kr == KERN_SUCCESS)
+
+        return Mach.Port<Mach.SendRight>(name:name)
+    }
+}
+
+
+extension Mach.Port where RightType == Mach.SendOnceRight {
+    /// Transfer ownership of the underlying port right to the caller.
+    ///
+    /// Returns the Mach port name representing the right.
+    ///
+    /// This operation liberates the right from management by the Mach.Port,
+    /// and the underlying right will no longer be automatically deallocated.
+    ///
+    /// After this function completes, the Mach.Port is destroyed and no longer
+    /// usable.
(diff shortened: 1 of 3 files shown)


Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+
+#if $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
+
+import Darwin.Mach
+
+protocol MachPortRight {}
+
+enum Mach {
+    @_moveOnly
+    struct Port<RightType:MachPortRight> {
```
mwwa, line 18: Style: All type spacing in this file should be `RightType: MachPortRight`, `var name: mach_port_name_t` etc.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+        ///
+        /// If the type of the right does not match the type T of Mach.Port<T>
+        /// being constructed, behavior is undefined.
+        ///
+        /// The underlying port right will be automatically deallocated at the
+        /// end of the Mach.Port instance's lifetime.
+        ///
+        /// This initializer makes a syscall to guard the right.
+        init(name:mach_port_name_t) {
+            assert(name != mach_port_name_t(MACH_PORT_NULL))
```
mwwa, line 35:
```suggestion
            precondition(name != mach_port_name_t(MACH_PORT_NULL))
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+            self.name = name
+
+            if (RightType.self == ReceiveRight.self) {
```
mwwa, line 38:
```suggestion
            if RightType.self == ReceiveRight.self {
```

Sources/System/MachPort.swift
mwwa: Most of the `assert` in this file seem should be `precondition`.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+            if name != 0xFFFFFFFF /* MACH_PORT_DEAD */ {
+                if RightType.self == ReceiveRight.self {
+                    // recv rights must be mod ref'ed instead of deallocated
+                    let kr = mach_port_unguard(mach_task_self_, name, context)
+                    assert(kr == KERN_SUCCESS)
+
+                    let kr2 = mach_port_mod_refs(mach_task_self_, name, MACH_PORT_RIGHT_RECEIVE, -1)
+                    assert(kr2 == KERN_SUCCESS)
+                } else {
+                    mach_port_deallocate(mach_task_self_, name)
```
mwwa, line 73: This branch needs to check the return value?

mwwa: Given how prevalent the mach_call + precondition pairing is in this file, it probably deserves pulling out into a private helper function?

Something like:
```swift
private func machPrecondition(
    _ expected: kern_return_t = KERN_SUCCESS,
    file: StaticString = #file,
    line: Int = #line,
    _ body: @autoclosure () -> kern_return_t
) {
    let kr = body()
    precondition(kr == expected, file: file, line: line)
}
```

Would transform most of these into:
```swift
machPrecondition(mach_port_deallocate(mach_task_self_, name))
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+    struct SendOnceRight : MachPortRight {}
+
+    /// Create a connected pair of rights, one receive, and one send.
+    ///
+    /// This function will abort if the rights could not be created.
+    /// Callers may assert that valid rights are always returned.
+    static func allocatePortRightPair() -> (Mach.Port<Mach.ReceiveRight>, Mach.Port<Mach.SendRight>) {
+        var name = mach_port_name_t(MACH_PORT_NULL)
+        let secret = mach_port_context_t(arc4random())
+        withUnsafeMutablePointer(to: &name) { name in
```
mwwa, line 109: Remove the `var name` and replace with something that lets you use `let`:

```suggestion
        let name = withUnsafeMutablePointer(to: &name) { name in
             var options = mach_port_options_t()
             options.flags = UInt32(MPO_INSERT_SEND_RIGHT);
             return withUnsafeMutablePointer(to: &options) { options in
                 let kr = mach_port_construct(mach_task_self_, options, secret, name)
                 assert(kr == KERN_SUCCESS)
                 return name
             }
         }
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+            var options = mach_port_options_t()
+            options.flags = UInt32(MPO_INSERT_SEND_RIGHT);
+            withUnsafeMutablePointer(to: &options) { options in
+                let kr = mach_port_construct(mach_task_self_, options, secret, name)
+                assert(kr == KERN_SUCCESS)
+            }
+        }
+        return (Mach.Port<Mach.ReceiveRight>(name: name, context: secret), Mach.Port<Mach.SendRight>(name: name))
```
mwwa, line 117:
```suggestion
        return (Port(name: name, context: secret), Port(name: name))
```
Should be sufficient here.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+    /// Mach port right into a Mach.Port by name.
+    ///
+    /// This initializer aborts if name is MACH_PORT_NULL.
+    ///
+    /// If the type of the right does not match the type T of Mach.Port<T>
+    /// being constructed, the behavior is undefined.
+    ///
+    /// The underlying port right will be automatically deallocated when
+    /// the Mach.Port object is destroyed.
+    init(name:mach_port_name_t, context:mach_port_context_t) {
```
mwwa, line 132:
```suggestion
    init(name: mach_port_name_t, `guard`: mach_port_context_t) {
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+    ///
+    /// Returns a tuple containing the Mach port name representing the right,
+    /// and the context value used to guard the right.
+    ///
+    /// This operation liberates the right from management by the Mach.Port,
+    /// and the underlying right will no longer be automatically deallocated.
+    ///
+    /// After this function completes, the Mach.Port is destroyed and no longer
+    /// usable.
+    __consuming func relinquish() -> (mach_port_name_t, mach_port_context_t) {
```
mwwa, line 164: At some point we'd talked about using labels here to improve the clarity of the return value:
```swift
__consuming func relinquish() -> (name: mach_port_name_t, `guard`: mach_port_context_t)
```

Would allow users to use
```swift
value.name
value.guard
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+    ///
+    /// This function will abort if the right could not be created.
+    /// Callers may assert that a valid right is always returned.
+    func makeSendOnceRight() -> Mach.Port<Mach.SendOnceRight> {
+        // send once rights do not coalesce
+        var kr:kern_return_t = KERN_FAILURE
+        var newRight:mach_port_name_t = mach_port_name_t(MACH_PORT_NULL)
+        var newRightType:mach_port_type_t = MACH_PORT_TYPE_NONE
+
+        withUnsafeMutablePointer(to: &newRight) { newRight in
```
mwwa, line 213: Same comment here about not using single-use `var` variables where we could formulate it as `let`

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+                kr = mach_port_extract_right(mach_task_self_, name, mach_msg_type_name_t(MACH_MSG_TYPE_MAKE_SEND_ONCE), newRight, newRightType)
+
+            }
+        }
+
+        // The value of newRight is validated by the Mach.Port initializer
+        assert(kr == KERN_SUCCESS)
+        assert(newRightType == MACH_MSG_TYPE_MOVE_SEND_ONCE)
+
+        return Mach.Port<Mach.SendOnceRight>(name:newRight)
```
mwwa, line 224:
```suggestion
        return Port(name: newRight)
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+    /// This function will abort if the right could not be created.
+    /// Callers may assert that a valid right is always returned.
+    func makeSendRight() -> Mach.Port<Mach.SendRight> {
+        let how = MACH_MSG_TYPE_MAKE_SEND
+
+        // send and recv rights are coalesced
+        let kr = mach_port_insert_right(mach_task_self_, name, name, mach_msg_type_name_t(how))
+        assert(kr == KERN_SUCCESS)
+
+        return Mach.Port<Mach.SendRight>(name:name)
```
mwwa, line 240:
```suggestion
        return Port(name: name)
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+                    let info = UnsafeMutableRawPointer(status).bindMemory(to: integer_t.self, capacity: 1)
+                    let kr = mach_port_get_attributes(mach_task_self_, name, MACH_PORT_RECEIVE_STATUS, info, size)
+                    assert(kr == KERN_SUCCESS);
+                }
+            }
+            return status.mps_mscount
+        }
+
+        set {
+            let kr = mach_port_set_mscount(mach_task_self_, name, newValue)
```
mwwa, line 261: What's the use case for `set`?

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+        let how = MACH_MSG_TYPE_COPY_SEND
+
+        // send rights are coalesced
+        let kr = mach_port_insert_right(mach_task_self_, name, name, mach_msg_type_name_t(how))
+        if kr == KERN_INVALID_CAPABILITY {
+            throw Mach.PortRightError.deadName
+        }
+        assert(kr == KERN_SUCCESS)
+
+        return Mach.Port<Mach.SendRight>(name:name)
```
mwwa, line 298:
```suggestion
        return Port(name: name)
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+#if $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
+
+import Darwin.Mach
+
+protocol MachPortRight {}
+
+enum Mach {
```
mwwa, line 16: This entire change is missing visibility keywords on things that seem to be expected to be public. (The default is `internal`)

mwwa (changes requested):

Sources/System/MachPort.swift
loffgren: The compiler doesn't let me use it for the `to:` argument in the same statement that declares it.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,318 @@
...
+        body(name, context)
+    }
+
+    /// Create a send-once right for a given receive right.
+    ///
+    /// This does not affect the makeSendCount of the receive right.
+    ///
+    /// This function will abort if the right could not be created.
+    /// Callers may assert that a valid right is always returned.
+    func makeSendOnceRight() -> Mach.Port<Mach.SendOnceRight> {
```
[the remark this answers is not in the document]
loffgren: Yes, but only from receive rights, because this interface makes running out of ipc space or uref overflow fatal, and invalid name/right mean that either the Mach.Port has not successfully managed the right or the contract with Mach.Port was violated (the right being destroyed out from under it).

Sources/System/MachPort.swift
loffgren: I fixed the whitespace, but kept this as `context` both here and in the new tuple label.

loffgren:
@swift-ci test

lorentey (approved):
I added some style nits and a few notes about potential calling convention gotchas (especially for the ABI stable incarnation of this code), but this looks good! 👍

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+#if $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
+
+import Darwin.Mach
+
+public protocol MachPortRight {}
```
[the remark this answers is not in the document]
milseman: If possible, that is if we know what comment to add, I'd prefer to land them with the PR so that we don't accidentally miss something when shipping. Alternatively, we could require that _everything_ has such a comment and the expand availability script will error for decls that don't have them, but I don't know if that's harder to do.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+
+private func machPrecondition(
```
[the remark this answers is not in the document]
milseman: There's also some historical baggage behind that rule in the stdlib that doesn't necessarily apply for System. I do agree it's a nice thing to do for module-wide (`internal`) visibility, but I don't know if I agree that little private functions and vars need them. This is not a strong opinion against them on my part, so much as a slight opposition to a strong opinion for them specifically for `private` entities.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+    _ body: @autoclosure () -> kern_return_t
+) {
+    let kr = body()
+    let expected = KERN_SUCCESS
+    precondition(kr == expected, file: file, line: line)
+}
+
+public enum Mach {
+    @_moveOnly
+    public struct Port<RightType: MachPortRight> {
```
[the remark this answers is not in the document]
milseman: The reason why most types in System are `@frozen` is because they are defined to be layout-compatible wrappers around C types (and often are even `RawRepresentable` with the C type). `FilePath` is not frozen because it's a more complex, higher level conceptual type and it's not a big deal to pass it indirectly.

Is there any downside to passing this around indirectly? Is there any desire to be able to reason about its layout by, e.g., having an Array of these in contiguous memory without indirection?

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+                    machPrecondition(mach_port_mod_refs(mach_task_self_, name, MACH_PORT_RIGHT_SEND, -1))
+                } else if RightType.self == SendOnceRight.self {
+                    machPrecondition(mach_port_mod_refs(mach_task_self_, name, MACH_PORT_RIGHT_SEND_ONCE, -1))
+                }
+            }
+        }
+    }
+
+    /// Possible errors that can be thrown by Mach.Port operations.
+    public enum PortRightError : Error {
```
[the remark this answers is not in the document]
milseman: I agree that in practice, it is often the case that either a `struct` or `@frozen enum` makes far more sense. How is this type meant to evolve over time? System tries pretty hard to stay away from enums that are not definitely exhaustive and definitely do not care whatsoever about layout. 

The only API enum is `FilePath.Component.Kind`, which is frozen, and is an exhaustive enumeration of kinds in which there is no layout to be concerned with.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+            options.flags = UInt32(MPO_INSERT_SEND_RIGHT);
+            withUnsafeMutablePointer(to: &options) { options in
+                machPrecondition(mach_port_construct(mach_task_self_, options, secret, name))
+            }
+        }
+        return (Mach.Port(name: name, context: secret), Mach.Port(name: name))
+    }
+}
+
+public extension Mach.Port where RightType == Mach.ReceiveRight {
```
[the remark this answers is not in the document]
milseman: Wait, does everything in here get public by default? I didn't even know that...

We do (I do at least) like having availability on the extension, though.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+    /// Borrow access to the port name in a block that can perform
+    /// non-consuming operations.
+    ///
+    /// Take care when using this function; many operations consume rights.
+    ///
+    /// If the right is consumed, behavior is undefined.
+    ///
+    /// The body block may optionally return something, which will then be
+    /// returned to the caller of withBorrowedName.
+    func withBorrowedName<ReturnType>(body: (mach_port_name_t, mach_port_context_t) -> ReturnType) -> ReturnType {
```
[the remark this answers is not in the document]
milseman: I'm not sure that it does strictly follow 80-col, but I do agree that the line wrapping shown is better here.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+        internal var context: mach_port_context_t
+
+        /// Transfer ownership of an existing unmanaged Mach port right into a
+        /// Mach.Port by name.
+        ///
+        /// This initializer aborts if name is MACH_PORT_NULL, or if name is
+        /// MACH_PORT_DEAD and the type T of Mach.Port<T> is Mach.ReceiveRight.
+        ///
+        /// If the type of the right does not match the type T of Mach.Port<T>
+        /// being constructed, behavior is undefined.
```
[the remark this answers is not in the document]
milseman: I agree to keep the term "undefined" aligned with C's notion of "undefined behavior", e.g. memory unsafety. What happens here, is it merely an unspecified result? For example, undefined might read from memory out of bounds while an unspecified result might throw errors when you try to use it without violating memory safety.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+
+ Copyright (c) 2022 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+#if $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
+
+import Darwin.Mach
```
milseman, line 12: Does this need to be `@_implementationOnly`? If so, does that forbid these types from being used in API without either our own compatible type aliases and/or `RawRepresentable` wrappers?

We try hard to keep System as a formal layer above `Darwin` without pulling a bunch of stuff into the global namespace. I don't think we want to cross this bridge here. @lorentey ?

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
+/*
+ This source file is part of the Swift System open source project
+
+ Copyright (c) 2022 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+#if $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
milseman, line 10: Does this need to have any interaction with required tools version for the package manifest, or does this just gracefully work out? @glessard do you know?

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+        /// The underlying port right will be automatically deallocated at the
+        /// end of the Mach.Port instance's lifetime.
+        ///
+        /// This initializer makes a syscall to guard the right.
+        public init(name: mach_port_name_t) {
+            precondition(name != mach_port_name_t(MACH_PORT_NULL), "Mach.Port cannot be initialized with MACH_PORT_NULL")
+            self.name = name
+
+            if RightType.self == ReceiveRight.self {
+                precondition(name != 0xFFFFFFFF /* MACH_PORT_DEAD */, "Receive rights cannot be dead names")
```
milseman, line 50: Out of curiosity, why do we need the constant value instead of the macro as in `MACH_PORT_NULL` above?

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+        /// If the right is consumed, behavior is undefined.
+        ///
+        /// The body block may optionally return something, which will then be
+        /// returned to the caller of withBorrowedName.
+        public func withBorrowedName<ReturnType>(body: (mach_port_name_t) -> ReturnType) -> ReturnType {
+            return body(name)
+        }
+
+        deinit {
+            if name == 0xFFFFFFFF /* MACH_PORT_DEAD */ {
```
milseman, line 76: If we use a constant twice, definitely hoist it and give it a name.

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+    /// Possible errors that can be thrown by Mach.Port operations.
+    public enum PortRightError : Error {
+        /// Returned when an operation cannot be completed, because the Mach
+        /// port right has become a dead name. This is caused by deallocation of the
+        /// receive right on the other end.
+        case deadName
+    }
+
+    /// The MachPortRight type used to manage a receive right.
+    public struct ReceiveRight : MachPortRight {}
```
milseman, line 100: Super tiny nit

```suggestion
    public struct ReceiveRight: MachPortRight {}
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+
+    /// The MachPortRight type used to manage a send right.
+    public struct SendRight : MachPortRight {}
```
milseman, line 103:
```suggestion
    public struct SendRight: MachPortRight {}
```

Sources/System/MachPort.swift
```diff
@@ -0,0 +1,327 @@
...
+
+    /// The MachPortRight type used to manage a send-once right.
+    ///
+    /// Send-once rights are the most restrictive type of Mach port rights.
+    /// They cannot create other rights, and are consumed upon use.
+    ///
+    /// Upon destruction a send-once notification will be sent to the
+    /// receiving end.
+    public struct SendOnceRight : MachPortRight {}
```
milseman, line 112:
```suggestion
    public struct SendOnceRight: MachPortRight {}
```

milseman (commented):
The `@_implementationOnly` part might throw a slight wrench into this.

Sources/System/MachPort.swift
loffgren: This is actually an issue with the Darwin Mach overlay, which has this name, but it doesn't work. From what I can tell, the Swift importer couldn't tell this should have been an unsigned value and exposed it as -1, which is not convertible to the unsigned `mach_port_name_t` type.

Sources/System/MachPort.swift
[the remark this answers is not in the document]
glessard: My initial reading triggered all the alarms for me because I thought it actually meant undefined behavior in the C sense, and we really don't want that. Now I simply don't know what the sentence means. It would be advantageous to clarify this as fully as possible!

Sources/System/MachPort.swift
glessard: The `#if $SomethingFeature` (magic) notation works as expected on every swift compiler, taking advantage of an unintentional loophole. It is supposed to be superseded by `#if hasFeature(somethingFeature)` in 5.8 and later.

glessard: Sorry, I was wrong. It only works since 5.3. It should be `#if compiler(>=5.3) && $MoveOnly`.

Sources/System/MachPort.swift
lorentey: I think this is fine! We initially misunderstood the semantics here -- publicly importing Darwin doesn't actually reexport its contents as a side effect. IIRC, at some point before the first release we realized this and switched it all back to regular imports. (Which allows us to refer to the original C types in the public API surface -- which would've been impossible to legally do otherwise.)

For example, the code below will not build unless the `import Darwin.Mach` statement gets uncommented:

```
import SystemPackage
//import Darwin.Mach

func t() {
    let _ = mach_port_name_t() // Cannot fine 'mach_port_name_t' in scope
}
``` 

So all is well!

(Importing `XCTest` or `Foundation` will, however, pollute local name lookup with symbols from `Darwin` -- IIRC, due to an unfortunate wrinkle of clang's module semantics. Oh well.)

Sources/System/MachPort.swift
gottesmm: So I went and asked Doug about this. The $MoveOnly syntax works back to the beginning of Swift as long as you are not using new syntax. If we use new syntax, then we need the compiler version there (this is because we still parse the code in the `#if` even if we are eventually going to throw it away). We aren't using new syntax here I think, so I don't think we need the compiler check. But if we eventually do introduce new syntax that can't be parsed by earlier compilers, then we would need the check.

lorentey: The `$` syntax is supposed to make this palatable to at least _some_ older Swift compilers; however, we really should bump the required tools version from 5.2 to at least 5.5 or 5.6 in the next feature release.

If there are any issues, they should show up during manual prerelease testing.

Note: We won't be able to ship a package release that includes `$MoveOnly` features until we have an actual compiler release that supports them. (We cannot consider noncopyable types to be a thing until then.)

Sources/System/MachPort.swift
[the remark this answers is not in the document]
lorentey: Yep, the access modifier on the `extension` gets applied by default to all members within. It's not the best.

Sadly we don't have much of a choice about putting `@available` only on the extension, as the compiler (and/or the ABI checker) will insist that it must be individually set on each member declaration. This isn't great, either.

Sources/System/MachPort.swift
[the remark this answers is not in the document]
lorentey: That would indeed be nice, but it'd be tricky to implement it in a barebones shell script. 😛
FWIW, recent compiler builds seem to produce warnings when encountering a declaration without an `@available` attribute while building ABI stable code, so we may get this for free!

Sources/System/MachPort.swift
lorentey: We have the option to add public `Mach.Port.null` and `Mach.Port.dead` names for these, as static computed properties that return `mach_port_name_t` values (or even `Mach.Port`, if that makes sense).

Sources/System/MachPort.swift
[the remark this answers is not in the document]
loffgren: The current interface is simple enough that you are most likely to hit unexpected traps for things that don't throw and are expected to always succeed for the correct type (in general every `machPrecondition`), but things may also silently succeed if your task does hold a right of that type at the time of the operation. However, some operations may have more complex failure modes, especially when the interface grows to encompass more complex things like messaging. e.g. If you created a `Mach.Port<Mach.SendRight>` with a send-once right, merely sending a message over that right consumes the right, which would cause downstream operations to trap.

lorentey:
@swift-ci test

loffgren:
@swift-ci test

lorentey:
Merging, to have the APIs in place. @loffgren is planning to follow up with additional work to resolve the outstanding comments -- they are all cosmetic/doc changes, or only relevant to the ABI stable incarnation, or recommending non-breaking additions. 

(Luckily, System importing C modules without `@_implementationOnly` is not going to be an issue!)

Sources/System/MachPort.swift
buttaface: Hmm, strangely [this broke building with the Swift 5.7.2 compiler, which I just noticed on my Android CI](https://github.com/buttaface/swift-android-sdk/actions/runs/3764473555/jobs/6398952221#step:17:25), though it works with the first 5.8 snapshot that was just released. I am able to reproduce the error when building natively with 5.7.2 for Ubuntu x86_64 too, so this will need to be modified, @loffgren. I notice the CI here only tests with trunk, not with the release 5.7 compiler, so presumably that's how this was missed.

PR #118, opened 2023-01-05
MachPort: fix build issues with Swift 5.7

### LINKED ISSUE #117: Trunk compilation is broken with the Swift 5.7 compiler on linux
I [reported this earlier on the offending pull](https://github.com/apple/swift-system/pull/116#discussion_r1056176682), but filing an issue to make sure it is handled.

The error:
```
Sources/System/MachPort.swift:75:5: error: deinitializers may only be declared within a class or actor
    deinit {
    ^
```

MaxDesiatov:
It seems that `#if $MoveOnly` check has no effect in Swift 5.7, so we should check whether Swift 5.8 or later are available first.

Resolves #117.

### DIFF
--- a/Sources/System/MachPort.swift
+++ b/Sources/System/MachPort.swift
@@ -7,7 +7,7 @@
  See https://swift.org/LICENSE.txt for license information
 */
 
-#if $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
+#if swift(>=5.8) && $MoveOnly && (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
 
 import Darwin.Mach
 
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

buttaface:
There was [some talk in the pull that added this that they wanted it to work back to Swift 5.5](https://github.com/apple/swift-system/pull/116/files/47feb93a38cc69af60411392977ac30e53fe7cf7#diff-3919bf8e0d2d560b4b749d6587bcc50a7485b6b5d27ea5738a99b94eb3b16b76), wouldn't this pull break that on Darwin?

MaxDesiatov:
I'm glad that 5.5 support is planned, my only goal for now is to fix the current breakage with 5.7 in the meantime.

buttaface:
Right, but won't this pull break building on 5.7 for macOS though?

MaxDesiatov:
@swift-ci test

weissi:
@lorentey friendly ping. Without this (and a tag), `swift-system` doesn't compile with any released version of Swift anymore. This includes actually tagged versions of `swift-system`.

PR #121, opened 2023-02-14
Mach.Port: Tidiness

loffgren:
- Add leading underscores where necessary
- Freeze the Mach namespace
- Mark Mach namespace available in System 1.3.0
- Fix 80 column signatures


loffgren:
@swift-ci test

lorentey (approved):
These are good improvements, thanks! 👍 

@karwa is right about `public extension` and the potential performance trap about non-`@inlinable` generic members.

loffgren:
@swift-ci test

PR #122, opened 2023-03-01
Mach: Avoid non-inlinable generics where possible

lorentey:
- Leave Ports as non-frozen for now, but expose its two stored properties.
- Make most members inlinable. The core initializers and the deinitializer must remain opaque for now.
- Mark the empty `MachPortRight` types as `@frozen` — these are markers, not carrying any data.

rdar://106057781


loffgren (approved):

lorentey:
CI isn't going to actually build or test this, sadly.

@swift-ci test

PR #131, opened 2023-06-21
Add availability attributes to `Sendable` conformances

glessard:
It is desirable to attach the same availability to the extensions declaring `Sendable` as to their base types.


glessard:
@swift-ci please test

lorentey (approved):
Adding the new conformances technically(*) requires a feature release, so an alternative would be to just get rid of compilers <5.5 and to add these directly to the type declarations, without having to repeat their availability.

(*) Then again, `Sendable` is special enough that it may be okay to add in a patch version.

PR #133, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.

### DIFF
--- a/Sources/System/Internals/CInterop.swift
+++ b/Sources/System/Internals/CInterop.swift
@@ -9,9 +9,12 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
-#elseif os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android)
+#elseif canImport(Glibc)
 @_implementationOnly import CSystem
 import Glibc
+#elseif canImport(Musl)
+@_implementationOnly import CSystem
+import Musl
 #elseif os(Windows)
 import CSystem
 import ucrt
--- a/Sources/System/Internals/CInterop.swift
+++ b/Sources/System/Internals/CInterop.swift
@@ -9,15 +9,15 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
+#elseif os(Windows)
+import CSystem
+import ucrt
 #elseif canImport(Glibc)
 @_implementationOnly import CSystem
 import Glibc
 #elseif canImport(Musl)
 @_implementationOnly import CSystem
 import Musl
-#elseif os(Windows)
-import CSystem
-import ucrt
 #else
 #error("Unsupported Platform")
 #endif
--- a/Sources/System/Internals/Constants.swift
+++ b/Sources/System/Internals/Constants.swift
@@ -458,7 +458,7 @@ internal var _O_RDWR: CInt { O_RDWR }
 
 #if !os(Windows)
 #if canImport(Musl)
-internal var _O_ACCMODE: CInt { 03|O_SEARCH }
+internal var _O_ACCMODE: CInt { 0x03|O_SEARCH }
 #else
 // TODO: API?
 @_alwaysEmitIntoClient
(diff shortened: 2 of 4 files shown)


Sources/System/Internals/Constants.swift
```diff
@@ -454,9 +457,13 @@ internal var _O_WRONLY: CInt { O_WRONLY }
 internal var _O_RDWR: CInt { O_RDWR }
 
 #if !os(Windows)
+#if canImport(Musl)
+internal var _O_ACCMODE: CInt { 03|O_SEARCH }
```
MaxDesiatov, line 461: Copied verbatim from the header since ClangImport doesn't support C macros that expand to expressions.

MaxDesiatov:
@swift-ci test

Sources/System/Internals/CInterop.swift
```diff
@@ -9,9 +9,12 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
-#elseif os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android)
+#elseif canImport(Glibc)
```
glessard, line 12: This is a big switch from OS name to capability. Would it make sense to move this to _after_ the `#elseif os(Windows)` branch?

MaxDesiatov:
@swift-ci test

Sources/System/Internals/CInterop.swift
glessard: We should do the same in the other files where there's a sequence of if-os and if-canImport.

Sources/System/Internals/Constants.swift
```diff
@@ -13,8 +13,11 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
-#elseif os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android)
+#elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import CSystem
```
tomerd, line 19: in other spots it is imported with @_implementationOnly, also, do we need this in the Glibc case as well?

Sources/System/Internals/Exports.swift
```diff
@@ -14,9 +14,12 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
-#elseif os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android)
+#elseif canImport(Glibc)
 @_implementationOnly import CSystem
 import Glibc
+#elseif canImport(Musl)
+import CSystem
```
tomerd, line 21: @_implementationOnly import CSystem?

Sources/System/Internals/CInterop.swift
```diff
@@ -9,12 +9,15 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
-#elseif os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android)
-@_implementationOnly import CSystem
-import Glibc
 #elseif os(Windows)
 import CSystem
 import ucrt
+#elseif canImport(Glibc)
```
tomerd, line 15: much better!

Sources/System/Internals/Constants.swift
MaxDesiatov: Not sure about this one, I'd stick to how other imports are done in this file for consistency.

MaxDesiatov:
@swift-ci test

Sources/System/Internals/Constants.swift
al45tair: Yuck. The original `03` is in _octal_. It so happens that 03<sub>8</sub> and 03<sub>10</sub> have the same value, but that's pretty horrid. I think I would _prefer_ that if we're doing to write this in Swift code, we explicitly write `0o03` or `0x03` rather than just `03`, because the latter _looks like_ octal if you're used to C family languages, but in Swift it is in fact decimal with a leading zero.

al45tair (approved):

MaxDesiatov:
@swift-ci test

PR #135, opened 2023-07-13
Tests: adjust the `FilePathTest.testFilePath` to build on Windows

compnerd:
UTF-8 is not the universal encoding for file systems.  Use a platform specific invalid string encoding for testing the sequence.  The internal representation of a `FilePath` is dependent on the system, and Windows uses UTF-16 as the base encoding for C interoperability.  Adjust the test to use an invalid sequence in the proper encoding.  This allows the tests to now build on Windows once again.


compnerd:
@swift-ci please test

al45tair (approved):

PR #136, opened 2023-07-13
Tests: avoid unnecessary duplication (NFCI)

compnerd:
Rather than specialising the function with the same implementation, prefer to use the `CInterop.PlatformUnicodeEncoding` to determine the correct code point encoding.  This allows us to share the implementation across the platforms.


compnerd:
@swift-ci please test

al45tair (approved):
Nice :-)

PR #139, opened 2023-08-07
fix _StrSlice == method linkage

### LINKED ISSUE #138: Build fails in release with: ld: Undefined symbols: static (extension in SystemPackage):SystemPackage._StrSlice.== infix(A, A) -> Swift.Bool
Any package which depends on `swift-system` and uses the `==` on `FilePath.Component` will fail to build _in release mode_ (debug is fine).

Consider this example program

```swift
import SystemPackage

print(FilePath("/x").lastComponent! == FilePath("/x").lastComponent!)
```

with the following dependencies/targets

```
    dependencies: [
        .package(url: "https://github.com/apple/swift-system.git", from: "1.0.0"),
    ],
    targets: [
        .executableTarget(
            name: "repro",
            dependencies: [
                .product(name: "SystemPackage", package: "swift-system"),
            ]
        ),
    ]
```

debug works

```
$ swift run 
Building for debugging...
Build complete! (0.34s)
true
```

release fails

```
$ swift run -c release
Building for production...
error: link command failed with exit code 1 (use -v to see invocation)
ld: Undefined symbols:
  static (extension in SystemPackage):SystemPackage._StrSlice.== infix(A, A) -> Swift.Bool, referenced from:
      _repro_main in main.swift.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: fatalError
[0/1] Linking repro
```

Also reproes in Xcode.

This is Xcode 15A221 but also reproduces on anything else I tried.

weissi:
A likely better alternative fix is in #140 

- fixes #138

### DIFF
--- a/Sources/System/FilePath/FilePath.swift
+++ b/Sources/System/FilePath/FilePath.swift
@@ -41,18 +41,18 @@
 public struct FilePath {
   // TODO(docs): Section on all the new syntactic operations, lexical normalization, decomposition,
   // components, etc.
-  internal var _storage: SystemString
+  internal var _storage: _SystemString
 
   /// Creates an empty, null-terminated path.
   public init() {
-    self._storage = SystemString()
+    self._storage = _SystemString()
     _invariantCheck()
   }
 
   // In addition to the empty init, this init will properly normalize
   // separators. All other initializers should be implemented by
   // ultimately deferring to a normalizing init.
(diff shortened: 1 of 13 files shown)


milseman (changes requested):
Let's go with the alternative that doesn't make the protocol public.

PR #140, opened 2023-08-07
alternative == method linkage fix

weissi:
alternative (to #139) fix for #138

### DIFF
--- a/Sources/System/FilePath/FilePathComponents.swift
+++ b/Sources/System/FilePath/FilePathComponents.swift
@@ -72,6 +72,16 @@ extension FilePath {
   }
 }
 
+extension FilePath.Component {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
+    return Self.strSliceEqual(lhs: lhs, rhs: rhs)
+  }
+
+  public func hash(into hasher: inout Hasher) {
+    return self.strSliceHash(into: &hasher)
+  }
+}
+
 @available(/*System 0.0.2: macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0*/iOS 8, *)
 extension FilePath.Component {
 
@@ -166,14 +176,21 @@ extension _StrSlice {
   internal var _systemString: SystemString { SystemString(_slice) }
 }
 extension _StrSlice {
-  public static func == (lhs: Self, rhs: Self) -> Bool {
+  public static func strSliceEqual(lhs: Self, rhs: Self) -> Bool {
     lhs._slice.elementsEqual(rhs._slice)
   }
-  public func hash(into hasher: inout Hasher) {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
+    return Self.strSliceEqual(lhs: lhs, rhs: rhs)
+  }
+
+  public func strSliceHash(into hasher: inout Hasher) {
     hasher.combine(_slice.count) // discriminator
     for element in _slice {
       hasher.combine(element)
     }
+  }
+  public func hash(into hasher: inout Hasher) {
+    return self.strSliceHash(into: &hasher)
   }
 }
 internal protocol _PathSlice: _StrSlice {


weissi:
@swift-ci please test

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -72,6 +72,16 @@ extension FilePath {
   }
 }
 
+extension FilePath.Component {
```
milseman, line 75:
```suggestion
extension FilePath.Component: Hashable {
```

@lorentey would this be possible to do (without availability) as the ABI-stable build picked up the symbols?

Also, would we still need the `public func` below if we added the conformance explicitly?

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -166,15 +176,22 @@ extension _StrSlice {
   internal var _systemString: SystemString { SystemString(_slice) }
 }
 extension _StrSlice {
-  public static func == (lhs: Self, rhs: Self) -> Bool {
+  public static func strSliceEqual(lhs: Self, rhs: Self) -> Bool {
```
lorentey, line 179: If we want to go this way, let's at least put an end to adding any new `public` members of internal types. That only leads to sadness.

```suggestion
  internal static func strSliceEqual(lhs: Self, rhs: Self) -> Bool {
```

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -72,6 +72,16 @@ extension FilePath {
   }
 }
 
+extension FilePath.Component {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
```
lorentey, line 76: Unfortunately, this would be a newly exported symbol, so it needs to come with an availability declaration.

To clean up the mess that resulted from adding `public` members to an `internal` type, we should add this implementation in a form that is back-deployable.

Sources/System/FilePath/FilePathComponents.swift
lorentey: The re-declaration of the conformance is probably fine, ABI-wise: this type already shipped with this.

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -72,6 +72,16 @@ extension FilePath {
...
+    return Self.strSliceEqual(lhs: lhs, rhs: rhs)
+  }
+
+  public func hash(into hasher: inout Hasher) {
```
lorentey, line 80: Same problem here -- this is a new symbol, so it needs to either have an `@available` attribute combined with `@backDeployed(before:)`, or it needs to be declared `@_alwaysEmitIntoClient`.

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -166,15 +176,22 @@ extension _StrSlice {
   internal var _systemString: SystemString { SystemString(_slice) }
 }
 extension _StrSlice {
-  public static func == (lhs: Self, rhs: Self) -> Bool {
+  public static func strSliceEqual(lhs: Self, rhs: Self) -> Bool {
     lhs._slice.elementsEqual(rhs._slice)
   }
-  public func hash(into hasher: inout Hasher) {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
```
lorentey, line 182: We should remove this function altogether. (After verifying that it doesn't end up getting exported somehow in the ABI stable binary.)

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -166,15 +176,22 @@ extension _StrSlice {
...
+    return Self.strSliceEqual(lhs: lhs, rhs: rhs)
+  }
+
+  public func strSliceHash(into hasher: inout Hasher) {
     hasher.combine(_slice.count) // discriminator
     for element in _slice {
       hasher.combine(element)
     }
   }
+  public func hash(into hasher: inout Hasher) {
```
lorentey, line 192: Same here -- this function should not exist.

Sources/System/FilePath/FilePathComponents.swift
```diff
@@ -166,15 +176,22 @@ extension _StrSlice {
...
-  public static func == (lhs: Self, rhs: Self) -> Bool {
+  public static func strSliceEqual(lhs: Self, rhs: Self) -> Bool {
     lhs._slice.elementsEqual(rhs._slice)
   }
-  public func hash(into hasher: inout Hasher) {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
+    return Self.strSliceEqual(lhs: lhs, rhs: rhs)
+  }
+
+  public func strSliceHash(into hasher: inout Hasher) {
```
lorentey, line 186:
```suggestion
  internal func strSliceHash(into hasher: inout Hasher) {
```

Sources/System/FilePath/FilePathComponents.swift
lorentey: The sinking of the Equatable/Hashable conformances from `_StrSlice` to `Component` is probably fine, ABI-wise. If any other public type conformed to `_StrSlice`, then it'll need to do the same.

lorentey:
I'll see if I can add a commit to this PR to make it work in all configurations of System later today!

PR #144, opened 2023-09-05
Fix typos in renaming message for `Errno`

kateinoigakukun:
There were some mismatches between the `renamed: ` suggestion and the actual variable names


kateinoigakukun:
@swift-ci please test

glessard (approved):
Thanks for catching this!

PR #152, opened 2024-02-20
Add support for visionOS

### LINKED ISSUE #150: Does not compile for visionOS
Compiling for visionOS in Xcode 15.2 on 23C71 yields an enormous number of build errors (180+)

noahpistilli:
Adds compilation support for visionOS. All tests pass. Closes #150.

I was unsure if I had to edit `Utilities/expand-availability.py`, so I didn't touch it.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -30,6 +30,7 @@ let package = Package(
           .define("_CRT_SECURE_NO_WARNINGS")
         ],
         swiftSettings: [
+          .define("SYSTEM_PACKAGE_DARWIN", .when(platforms: [.macOS, .iOS, .watchOS, .tvOS, .visionOS])),
           .define("SYSTEM_PACKAGE"),
           .define("ENABLE_MOCKING", .when(configuration: .debug))
         ]),
@@ -37,6 +38,7 @@ let package = Package(
         name: "SystemTests",
         dependencies: ["SystemPackage"],
         swiftSettings: [
+          .define("SYSTEM_PACKAGE_DARWIN", .when(platforms: [.macOS, .iOS, .watchOS, .tvOS, .visionOS])),
           .define("SYSTEM_PACKAGE")
         ]),
     ]
--- a/Sources/System/Errno.swift
+++ b/Sources/System/Errno.swift
@@ -23,7 +23,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   private init(_ raw: CInt) { self.init(rawValue: raw) }
 
-#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+#if SYSTEM_PACKAGE_DARWIN
   /// Error. Not used.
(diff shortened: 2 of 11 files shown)


glessard:
I would prefer to do this by setting a compilation variable in the package manifest. This `os||os||os||os||os` list was already unwieldy, and it's only getting worse.

noahpistilli:
Perhaps something like
```swift
swiftSettings: [
      .define("DARWIN", .when(platforms: [Platform.macOS, Platform.iOS, Platform.watchOS, Platform.tvOS, Platform.visionOS])),
```

in the package manifest?

rauhul:
> Perhaps something like
> 
> ```swift
> 
> swiftSettings: [
> 
>       .define("APPLE", .when(platforms: [Platform.macOS, Platform.iOS, Platform.watchOS, Platform.tvOS, Platform.visionOS])),
> 
> ```
> 
> 
> 
> in the package manifest?

If you do go this route note that Apple has non Darwin uses of Swift such as the triple: `armv7em-apple-none-macho`. 

You could name this define `DARWIN` instead?

glessard (approved):
This is great, thanks for helping with this!

glessard:
@swift-ci please test

glessard:
The testing failure involves a change in Swift 6.0. Will need to fix this separately.

glessard:
@noahpistilli I think the approach is good, but the symbol "DARWIN" may be a bit too prone to collisions, given that it could possibly be passed from the command-line by any package that uses this. May I suggest something less likely, such as "SYSTEM_PACKAGE_DARWIN"?

glessard:
@swift-ci please test

karwa:
Wouldn't it be better to use `canImport(Darwin)`? 

Especially since a lot of these are literally guarding imports of the `Darwin` module and uses of symbols exported by it...

rauhul:
> Wouldn't it be better to use `canImport(Darwin)`? 
> 
> 
> 
> Especially since a lot of these are literally guarding imports of the `Darwin` module and uses of symbols exported by it...

This seems abundantly obvious now that you say it...

glessard:
@swift-ci please test

glessard:
`#if canImport()` has a namespacing issue; nothing particular prevents a Linux distribution from having a module named Darwin. We've avoided using that approach for that reason.

karwa:
> `#if canImport()` has a namespacing issue; nothing particular prevents a Linux distribution from having a module named Darwin. We've avoided using that approach for that reason.

It’s only an issue if this package were to introduce a dependency on another package named “Darwin”, or if a third-party toolchain were to include a Swift module named “Darwin” in its standard installation.

That’s sufficiently unlikely that I don’t think it needs to be a concern. Such a toolchain would quickly encounter widespread problems across the Swift ecosystem and would likely choose a different name.

Importantly, the official toolchain building would not encounter issues when building for Apple platforms. That’s the only thing I think we need to be concerned about right now.

Issue #153, opened 2024-03-07
Fix `SystemString`'s `Collection` conformance


glessard:
Swift 6.0 has an improved `Array.replaceSubrange()` implementation which is stricter than before with its Collection parameters.

SystemString’s `withContiguousStorageIfAvailable()` passes an UnsafeBufferPointer whose count is larger than itself, to accommodate a trailing null byte.

Array.replaceSubrange() does not like that at all, and swift-system’s tests crash with pre-release Swift 6 toolchains.

rdar://124102547

PR #154, opened 2024-03-08
Fix `SystemString.withContiguousStorageIfAvailable()`

glessard:
- The original implementation simply forwarded to the underlying `nullTerminatedStorage`, which contains one element – a null terminator – more than the actual `String`.
- This isn’t in keeping with the sperit of `withContiguousStorageIfAvailable()`, which should present exactly the elements of the collection.

Addresses https://github.com/apple/swift-system/issues/153 (rdar://124102547)

### DIFF
--- a/Sources/System/SystemString.swift
+++ b/Sources/System/SystemString.swift
@@ -142,31 +142,38 @@ extension SystemString: RangeReplaceableCollection {
     nullTerminatedStorage.reserveCapacity(1 + n)
   }
 
-  // TODO: Below include null terminator, is this desired?
-
   internal func withContiguousStorageIfAvailable<R>(
     _ body: (UnsafeBufferPointer<SystemChar>) throws -> R
   ) rethrows -> R? {
-    try nullTerminatedStorage.withContiguousStorageIfAvailable(body)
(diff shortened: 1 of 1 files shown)


glessard:
@swift-ci please test

Sources/System/SystemString.swift
```diff
@@ -142,31 +142,38 @@ extension SystemString: RangeReplaceableCollection {
...
   }
 }
 
 extension SystemString: Hashable, Codable {}
 
 extension SystemString {
-  // TODO: Below include null terminator, is this desired?
 
+  // withSystemChars includes the null terminator
   internal func withSystemChars<T>(
```
lorentey, line 173: We should probably call this out in the name (`withNullTerminatedSystemChars` or something)

lorentey (approved):

Sources/System/SystemString.swift
glessard: I'll include that change along with a consistency change for `withCodeUnits`.

glessard: Suggested name change included in https://github.com/apple/swift-system/pull/155

PR #157, opened 2024-03-20
Update README.md

Coeur:


rauhul:
The use of "currency" here is intentional. It implies that these are not bespoke types just for swift-system, they are intended to be used by everyone and values of these types should be freely passed between libraries.

PR #158, opened 2024-03-20
Fix swift-system to work on Windows.

al45tair:
The swift-system tests were not passing, for various reasons, mostly to do with path related confusion.

As part of this, add a function to create a temporary directory, as that was one of the problems.

Also add Windows support for pipe() and ftruncate(), which was missing.

rdar://125087707

### DIFF
--- a/Sources/System/FilePath/FilePathParsing.swift
+++ b/Sources/System/FilePath/FilePathParsing.swift
@@ -89,6 +89,11 @@ extension SystemString {
       // `_prenormalizeWindowsRoots` and resume.
       readIdx = _prenormalizeWindowsRoots()
       writeIdx = readIdx
+
+      // Skip redundant separators
+      while readIdx < endIndex && isSeparator(self[readIdx]) {
+          self.formIndex(after: &readIdx)
+      }
     } else {
       assert(genericSeparator == platformSeparator)
     }
@@ -330,10 +335,13 @@ extension FilePath {
 // Whether we are providing Windows paths
 @inline(__always)
 internal var _windowsPaths: Bool {
+  if let forceWindowsPaths = forceWindowsPaths {
+    return forceWindowsPaths
+  }
   #if os(Windows)
   return true
   #else
-  return forceWindowsPaths
+  return false
   #endif
 }
 
--- a/Sources/System/FilePath/FilePathTemp.swift
+++ b/Sources/System/FilePath/FilePathTemp.swift
@@ -111,7 +111,33 @@ internal func recursiveRemove(at path: FilePath) throws {
 
 #else
 internal func getTemporaryDirectory() throws -> FilePath {
+  #if SYSTEM_PACKAGE_DARWIN
+  var capacity = 1024
+  while true {
+    let path: FilePath? = withUnsafeTemporaryAllocation(
+      of: CInterop.PlatformChar.self,
+      capacity: 1024) { buffer in
+      let len = system_confstr(SYSTEM_CS_DARWIN_USER_TEMP_DIR,
+                               buffer.baseAddress!,
+                               buffer.count)
+      if len == 0 {
+        // Fall back to "/tmp" if we can't read the temp directory
+        return "/tmp"
+      }
+      // If it was truncated, increase capaciy and try again
+      if len > buffer.count {
+        capacity = len
+        return nil
+      }
+      return FilePath(SystemString(platformString: buffer.baseAddress!))
+    }
+    if let path = path {
+      return path
+    }
+  }
+  #else
   return "/tmp"
+  #endif
 }
 
 internal func recursiveRemove(at path: FilePath) throws {
--- a/Sources/System/FilePath/FilePathTemp.swift
+++ b/Sources/System/FilePath/FilePathTemp.swift
@@ -24,7 +24,7 @@ public func withTemporaryPath<R>(basename: FilePath.Component,
 #if os(Windows)
 import WinSDK
 
-internal func getTemporaryDirectory() throws -> FilePath {
+fileprivate func getTemporaryDirectory() throws -> FilePath {
   return try withUnsafeTemporaryAllocation(of: CInterop.PlatformChar.self,
                                            capacity: Int(MAX_PATH) + 1) {
     buffer in
@@ -37,7 +37,7 @@ internal func getTemporaryDirectory() throws -> FilePath {
   }
 }
 
-internal func forEachFile(
+fileprivate func forEachFile(
   at path: FilePath,
   _ body: (WIN32_FIND_DATAW) throws -> ()
 ) rethrows {
@@ -67,7 +67,7 @@ internal func forEachFile(
   }
 }
 
-internal func recursiveRemove(at path: FilePath) throws {
+fileprivate func recursiveRemove(at path: FilePath) throws {
   // First, deal with subdirectories
   try forEachFile(at: path) { findData in
     if (findData.dwFileAttributes & DWORD(FILE_ATTRIBUTE_DIRECTORY)) != 0 {
@@ -109,13 +109,14 @@ internal func recursiveRemove(at path: FilePath) throws {
 }
 
 #else
-internal func getTemporaryDirectory() throws -> FilePath {
+fileprivate func getTemporaryDirectory() throws -> FilePath {
   #if SYSTEM_PACKAGE_DARWIN
   var capacity = 1024
   while true {
     let path: FilePath? = withUnsafeTemporaryAllocation(
       of: CInterop.PlatformChar.self,
-      capacity: capacity) { buffer in
+      capacity: capacity
+    ) { buffer in
       let len = system_confstr(SYSTEM_CS_DARWIN_USER_TEMP_DIR,
                                buffer.baseAddress!,
                                buffer.count)
@@ -139,7 +140,7 @@ internal func getTemporaryDirectory() throws -> FilePath {
   #endif
 }
 
-internal func recursiveRemove(at path: FilePath) throws {
+fileprivate func recursiveRemove(at path: FilePath) throws {
   let dirfd = try FileDescriptor.open(path, .readOnly, options: .directory)
   defer {
     try? dirfd.close()
@@ -173,7 +174,7 @@ fileprivate func impl_opendirat(
   return system_fdopendir(fd)
 }
 
-internal func forEachFile(
+fileprivate func forEachFile(
   in dirfd: CInt, path: UnsafePointer<CInterop.PlatformChar>,
   _ body: (system_dirent) throws -> ()
 ) throws {
@@ -238,11 +239,11 @@ internal func recursiveRemove(
 }
 #endif
 
-internal let base64 = Array<UInt8>(
+fileprivate let base64 = Array<UInt8>(
   "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_".utf8
 )
 
-internal func makeDirectory(at: FilePath) throws -> Bool {
+fileprivate func makeTempDirectory(at: FilePath) throws -> Bool {
   return try at.withPlatformString {
     if system_mkdir($0, 0o700) == 0 {
       return true
@@ -256,7 +257,7 @@ internal func makeDirectory(at: FilePath) throws -> Bool {
   }
 }
 
-internal func createRandomString(length: Int) -> String {
+fileprivate func createRandomString(length: Int) -> String {
   return String(
     decoding: (0..<length).map{
       _ in base64[Int.random(in: 0..<64)]
@@ -274,7 +275,7 @@ internal func createUniqueTemporaryDirectory(
   while true {
     tempDir.extension = createRandomString(length: 16)
 
-    if try makeDirectory(at: tempDir) {
+    if try makeTempDirectory(at: tempDir) {
       return tempDir
     }
   }
--- a/Sources/System/FilePath/FilePathTemp.swift
+++ b/Sources/System/FilePath/FilePathTemp.swift
@@ -1,0 +1,256 @@
+/*
+ This source file is part of the Swift System open source project
+
+ Copyright (c) 2024 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+// MARK: - API
+
+public func withTemporaryPath<R>(basename: FilePath.Component,
+                                 _ body: (FilePath) throws -> R) throws -> R {
+  let temporaryDir = try createUniqueTemporaryDirectory(basename: basename)
+  defer {
+    try? recursiveRemove(at: temporaryDir)
+  }
+
+  return try body(temporaryDir)
+}
+
+// MARK: - Internals
+
+#if os(Windows)
+import WinSDK
+
+internal func getTemporaryDirectory() throws -> FilePath {
+  return try withUnsafeTemporaryAllocation(of: CInterop.PlatformChar.self,
+                                           capacity: Int(MAX_PATH) + 1) {
+    buffer in
+
+    let count = GetTempPath2W(DWORD(buffer.count), buffer.baseAddress)
+    if count == 0 {
+      throw Errno(windowsError: GetLastError())
+    }
+
+    return FilePath(SystemString(platformString: buffer.baseAddress!))
+  }
+}
+
+internal func forEachFile(
+  at path: FilePath,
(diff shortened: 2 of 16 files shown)


al45tair:
@swift-ci Please test Windows platform

al45tair:
I guess we don't test Windows for swift-system. We should.

Sources/System/FilePath/FilePathParsing.swift
```diff
@@ -89,6 +89,11 @@ extension SystemString {
       // `_prenormalizeWindowsRoots` and resume.
       readIdx = _prenormalizeWindowsRoots()
       writeIdx = readIdx
+
+      // Skip redundant separators
+      while readIdx < endIndex && isSeparator(self[readIdx]) {
+          self.formIndex(after: &readIdx)
+      }
```
compnerd, line 96: This is kinda tricky, the duplicates are material at the beginning as the separator makes parsing differences. Internal duplicates are fine to normalize.

Sources/System/FilePath/FilePathTemp.swift
```diff
@@ -0,0 +1,256 @@
...
+#if os(Windows)
+import WinSDK
+
+internal func getTemporaryDirectory() throws -> FilePath {
+  return try withUnsafeTemporaryAllocation(of: CInterop.PlatformChar.self,
+                                           capacity: Int(MAX_PATH) + 1) {
+    buffer in
+
+    let count = GetTempPath2W(DWORD(buffer.count), buffer.baseAddress)
+    if count == 0 {
```
compnerd, line 33: I'd collapse this with a `guard`.

Sources/System/Internals/Mocking.swift
```diff
@@ -110,8 +110,8 @@ private var contextualMockingEnabled: Bool {
 extension MockingDriver {
   internal static var enabled: Bool { mockingEnabled }
 
-  internal static var forceWindowsPaths: Bool {
-    currentMockingDriver?.forceWindowsSyntaxForPaths ?? false
+  internal static var forceWindowsPaths: Optional<Bool> {
```
compnerd, line 113: Why the different spelling here? `Optional<Bool>` vs `Bool?` above.

Sources/System/Internals/Mocking.swift
```diff
@@ -128,7 +128,7 @@ internal var mockingEnabled: Bool {
 }
 
 @inline(__always)
-internal var forceWindowsPaths: Bool {
+internal var forceWindowsPaths: Optional<Bool> {
```
compnerd, line 131: Similar

Sources/System/Internals/Syscalls.swift
```diff
@@ -118,20 +118,90 @@ internal func system_dup2(_ fd: Int32, _ fd2: Int32) -> Int32 {
...
 }
+
+internal func system_mkdir(
+    _ path: UnsafePointer<CInterop.PlatformChar>,
+    _ mode: CInterop.Mode
+) -> CInt {
+#if ENABLE_MOCKING
+  if mockingEnabled { return _mock(path: path, mode) }
+#endif
+  return mkdir(path, mode)
```
compnerd, line 142: Is this shared? I think that we should prefer `CreateDirectoryW` over `mkdir`.

Sources/System/Internals/Syscalls.swift
```diff
@@ -118,20 +118,90 @@ internal func system_dup2(_ fd: Int32, _ fd2: Int32) -> Int32 {
...
+}
+
+internal func system_rmdir(
+    _ path: UnsafePointer<CInterop.PlatformChar>
+) -> CInt {
+#if ENABLE_MOCKING
+  if mockingEnabled { return _mock(path: path) }
+#endif
+  return rmdir(path)
```
compnerd, line 151: Similar, `RemoveDirectoryW` should be preferred.

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -115,11 +114,133 @@ internal func pwrite(
...
+
+  return 0;
+}
+
+@inline(__always)
+internal func mkdir(
+  _ path: UnsafePointer<CInterop.PlatformChar>,
+  _ mode: CInterop.Mode
+) -> CInt {
+  return _wmkdir(path)
```
compnerd, line 163: I think we should prefer `CreateDirectoryW`. The win32 API is safer.

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -115,11 +114,133 @@ internal func pwrite(
...
+}
+
+@inline(__always)
+internal func rmdir(
+  _ path: UnsafePointer<CInterop.PlatformChar>
+) -> CInt {
+  return _wrmdir(path)
```
compnerd, line 170: Similar, prefer the win32 api (`RemoveDirectoryW`).

Sources/System/FilePath/FilePathParsing.swift
al45tair: Yes. This skips duplicates _after_ the beginning has been processed (that's what the line above is doing). The problem was that we weren't skipping duplicates immediately after what this code refers to as the "root" (i.e. after UNC paths or drive letters).

Sources/System/Internals/Syscalls.swift
al45tair: This is shared. We're actually using `_wmkdir` for Windows at present.

Sources/System/Internals/Syscalls.swift
al45tair: Somewhat asymmetrically, we're using `RemoveDirectoryW` on Windows (see `FilePathTemp.swift`), mostly because the code for `recursiveRemove` isn't shared whereas `makeDirectory` is. Though maybe `makeDirectory` should use `CreateDirectoryW` on Windows and set a security descriptor?

al45tair:
@swift-ci Please test

Sources/System/FilePath/FilePathTemp.swift
```diff
@@ -0,0 +1,281 @@
...
+      // If it was truncated, increase capaciy and try again
+      if len > buffer.count {
+        capacity = len
+        return nil
+      }
+      return FilePath(SystemString(platformString: buffer.baseAddress!))
+    }
+    if let path = path {
+      return path
+    }
```
lorentey, line 135: This seems _far_ too much logic for swift-system. This package is not supposed to be the place for building cross-platform abstractions.

Sources/System/FilePath/FilePathTemp.swift
```diff
@@ -0,0 +1,281 @@
...
+ Copyright (c) 2024 Apple Inc. and the Swift System project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+*/
+
+// MARK: - API
+
+public func withTemporaryPath<R>(basename: FilePath.Component,
+                                 _ body: (FilePath) throws -> R) throws -> R {
```
lorentey, line 13: Nit: please follow the [stdlib's line breaking conventions](https://github.com/apple/swift/blob/main/docs/StandardLibraryProgrammersManual.md#line-breaking). 

```suggestion
public func withTemporaryPath<R>(
  basename: FilePath.Component,
  _ body: (FilePath) throws -> R
) throws -> R {
```

I cannot help questioning if this really needs to be a top-level function, rather than some static method on `FilePath`.

But more importantly: does this _really_ belong in swift-system?

Sources/System/FilePath/FilePathTemp.swift
```diff
@@ -0,0 +1,281 @@
...
+}
+#endif
+
+internal let base64 = Array<UInt8>(
+  "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_".utf8
+)
+
+internal func makeDirectory(at: FilePath) throws -> Bool {
+  return try at.withPlatformString {
+    if system_mkdir($0, 0o700) == 0 {
```
lorentey, line 247: If we want this function to create directories with the hardwired mode 700, then it should probably carry a more specific name than just `makeDirectory`.

Sources/System/FilePath/FilePathTemp.swift
```diff
@@ -0,0 +1,281 @@
...
+}
+
+internal func createUniqueTemporaryDirectory(
+  basename: FilePath.Component
+) throws -> FilePath {
+  var tempDir = try getTemporaryDirectory()
+  tempDir.append(basename)
+
+  while true {
+    tempDir.extension = createRandomString(length: 16)
```
lorentey, line 275: On POSIX systems, shouldn't we just call `mkdtemp` to make a unique temp directory?

lorentey (commented):
Yay for the FilePath fixes! 🎉 Thanks 👍 

I have lots of grumpy comments about the additions. swift-system was never intended to be a POSIX emulation package, and I really dislike that it's being pushed to become that. (I don't think we have the means to do a good job of that, even if it was a good idea, which I don't think it is.) This package isn't for building cross-platform abstraction layers -- that's Foundation's job.

Sources/System/FilePath/FilePathTemp.swift
```diff
@@ -0,0 +1,281 @@
...
+        return nil
+      }
+      return FilePath(SystemString(platformString: buffer.baseAddress!))
+    }
+    if let path = path {
+      return path
+    }
+  }
+  #else
+  return "/tmp"
```
lorentey, line 138: On POSIX systems, I think we should be looking at the `TMPDIR` envvar; we should not be hardwiring /tmp.

Sources/System/FilePath/FilePathTemp.swift
al45tair: `mkdtemp` is an especially, egregiously horrible function. It writes to its argument, which sadly often works with string "constants" in C because they historically have often ended up in writeable memory. I agree that it's a standard function that does something very much like this. It's also awful.

Sources/System/FilePath/FilePathTemp.swift
al45tair: Yes, that's a fair point, though I'm not certain getting the environment variables from `swift-system` is as easy as it should be. `getenv("TMPDIR")` in Swift doesn't appear to work for some reason I don't fully understand, and we don't want to be using Foundation from here.

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -115,11 +114,144 @@ internal func pwrite(
...
-    // TODO(compnerd) map windows error to errno
+    ucrt._set_errno(mapWindowsErrorToErrno(GetLastError()))
     return Int(-1)
   }
   return Int(nNumberOfBytesWritten)
 }
 
+@inline(__always)
+internal func pipe(_ fds: UnsafeMutablePointer<Int32>) -> CInt {
+  return _pipe(fds, 4096, _O_BINARY | _O_NOINHERIT);
```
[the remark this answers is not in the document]
al45tair: Playing devil's advocate for a moment, the `_pipe()` API provided by the Windows C library is _very_ close to the `pipe()` API provided by POSIX. Yes, it has two extra arguments; the first is the buffer size (on POSIX, you get whatever the kernel decides, so hard-coding it on Windows is hardly a huge issue), and the second is some flags that let you set text mode or handle inheritance (neither of which POSIX really has). I wouldn't call this "emulation" so much as calling the equivalent function.

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -115,11 +114,144 @@ internal func pwrite(
...
+}
+
+@inline(__always)
+internal func ftruncate(_ fd: Int32, _ length: off_t) -> Int32 {
```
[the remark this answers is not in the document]
al45tair: The entire C library on Windows is in some sense "emulated", since the C library is not the native platform API there. I think this likely explains the difference in approach from Saleem in the past. And yes, the UCRT doesn't have `ftruncate()` or any equivalent (that I'm aware of), but the Win32 API totally does all the things we need it to do to make that work, and it isn't terribly complicated to implement.

Note: I'm _not_ arguing for one moment that `swift-system` should become a POSIX emulation layer. Just observing that in practice, non-POSIX _largely_ means Windows at this point, and if anyone is using `FileDescriptor` for anything then they'll end up writing things that aren't portable to Windows unless we make sure that we have equivalent functionality. I don't think it makes sense to push yet more `#if os(Windows)` stuff up the stack.

(I'm also playing Devil's advocate here.)

al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test

compnerd (approved):

Sources/System/FilePath/FilePathTempWindows.swift
```diff
@@ -0,0 +1,114 @@
...
+/// - Parameters:
+///   - path: The directory to be deleted.
+///
+/// Removes a directory completely, including all of its contents.
+internal func _recursiveRemove(
+  at path: FilePath
+) throws {
+  // First, deal with subdirectories
+  try forEachFile(at: path) { findData in
+    if (findData.dwFileAttributes & DWORD(FILE_ATTRIBUTE_DIRECTORY)) != 0 {
```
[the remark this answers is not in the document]
al45tair: The file attribute constants are [documented to be a bit mask](https://learn.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants). I think the `!= 0` and `== 0` is idiomatic for Win32 code — not least because the alternative requires that the compiler optimise away an unnecessary comparison.

al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test Linux platform

al45tair:
@swift-ci Please test macOS platform

al45tair:
@swift-ci Please test Windows platform

Sources/System/FilePath/FilePathTemp.swift
```diff
@@ -0,0 +1,97 @@
...
+/// Create a temporary path for the duration of the closure.
+///
+/// - Parameters:
+///   - basename: The base name for the temporary path.
+///   - body: The closure to execute.
+///
+/// Creates a temporary directory with a name based on the given `basename`,
+/// executes `body`, passing in the path of the created directory, then
+/// deletes the directory and all of its contents before returning.
+public func withTemporaryFilePath<R>(
```
milseman, line 21: Make sure that any API additions go through some kind of review, ideally on the forums. At the very least, a dedicated PR that highlights it's an API addition PR as opposed to just a bug-fix PR for Windows.

For this function, would a static method on `FilePath` work instead of a global? E.g. `FilePath.withTemporaryPath { ... }`.

Sources/System/Internals/WindowsSyscallAdapters.swift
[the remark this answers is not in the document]
lorentey: Anything importing SystemPackage is expected to end up having to deal with `#if os(Foo)` conditions, anyway. 

SystemPackage is not designed to provide cross-platform API.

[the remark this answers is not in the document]
al45tair: Sure. We spoke about this on Slack… fundamentally the issue here is that from a Windows developer's standpoint, a (kind of) subset of the POSIX file descriptor API does exist on Windows and is useful for porting tools written for UNIX. It isn't all that useful for Windows programming more generally — Windows developers would generally drop down to the Win32 layer rather than using `_open()` et al — everyone knows that it's all emulated anyway.

Basically, I don't think it's worth _not_ supporting things that are trivially easy to implement, like `pipe()` and `ftruncate()`. I also don't think it should evolve into a Cygwin like layer, because that's _way_ too far.

I thought we'd agreed that this was the right thing for Windows; maybe I misunderstood the conversation we had?

Sources/System/FilePath/FilePathTemp.swift
al45tair: Yes, it would. My only reservation is that (to me) things like `FilePath.withTemporaryPath` read like a function that _constructs and returns_ a `FilePath` from some kind of `TemporaryPath` thing. This isn't the same as e.g. `Array`'s `withUnsafeBytes`, because that's called on an instance rather than on the type itself. Maybe that's just because I'm used to ObjC. It's obvious from the arguments that it takes a closure, of course.

I only needed this for the tests, so for now I think the thing to do is to make this non-public (I'll raise a PR to fix that). I'll check through to make sure I didn't leak any other new API and do the same there.

PR #161, opened 2024-05-02
Disfavor String.init(platformString: UnsafePointer<CChar>)

jmschonfeld:
This change disfavors `String(platformString:)` so that it is not ambiguous with `String(cString:)` when the initializer is specified without noting the label (i.e. in usages such as `someUnsafePointer.map(String.init)`

Resolves rdar://127408859


jmschonfeld:
@swift-ci please test

glessard:
@swift-ci please test

glessard (commented):
It's unfortunate that we need this, but it does resolve the problem.

PR #169, opened 2024-05-10
Improved compatibility for visionOS support

glessard:
The original version of visionOS support on main accidentally left out versions of the Swift compiler which aren't aware of the visionOS platform. Until we only support compilers that know the `.visionOS` platform, we need to be more accommodating.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,6 +12,13 @@
 
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


glessard:
@swift-ci please test

lorentey (approved):

PR #170, opened 2024-05-10
Make a platform-specific flag explicitly platform-specific

glessard:
It is more self-documenting to make it platform-specific.


glessard:
@swift-ci please test

Azoy (approved):

PR #175, opened 2024-05-12
fix a 32-bit overflow issue

glessard:
- ensures compatibility with watchOS


glessard:
@swift-ci please test

PR #176, opened 2024-05-13
Remove unintentional public functions and fix a couple of Windows issues.

al45tair:
A few things were marked `public` that should not have been, so change those to `internal`.  That means we then have to fix the tests to use `@testable` imports, so do that too.

Finally, while testing this, I noticed that there was a bug in the Windows code that meant that files opened for read-only access would open, but were not in fact readable (because we weren't asking for read permission, because `_O_RDONLY` happens to be zero).  Fix that also.

### DIFF
--- a/Sources/System/Internals/WindowsSyscallAdapters.swift
+++ b/Sources/System/Internals/WindowsSyscallAdapters.swift
@@ -621,14 +621,15 @@ fileprivate struct DecodedOpenFlags {
     }
 
     dwDesiredAccess = 0
-    if (oflag & _O_RDONLY) != 0 {
+    switch (oflag & (_O_RDONLY|_O_WRONLY|_O_RDWR)) {
+    case _O_RDONLY:
       dwDesiredAccess |= DWORD(GENERIC_READ)
-    }
-    if (oflag & _O_WRONLY) != 0 {
+    case _O_WRONLY:
       dwDesiredAccess |= DWORD(GENERIC_WRITE)
-    }
-    if (oflag & _O_RDWR) != 0 {
+    case _O_RDWR:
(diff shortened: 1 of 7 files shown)


al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test

al45tair:
Windows tests:
```
C:\Users\Alastair\Source\swift-system>swift test
...
Test Suite 'debug.xctest' passed at 2024-05-13 15:49:41.456
Executed 49 tests, with 0 failures (0 unexpected) in 5.453 (5.453) seconds
Test Suite 'All tests' passed at 2024-05-13 15:49:41.456
         Executed 49 tests, with 0 failures (0 unexpected) in 5.453 (5.453) seconds
```

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -621,14 +621,15 @@ fileprivate struct DecodedOpenFlags {
     }
 
     dwDesiredAccess = 0
-    if (oflag & _O_RDONLY) != 0 {
+    switch (oflag & (_O_RDONLY|_O_WRONLY|_O_RDWR)) {
```
compnerd, line 624: This probably deserves a comment to indicate that the flags are non-overlapping.

compnerd (approved):

Sources/System/Internals/WindowsSyscallAdapters.swift
al45tair: Done.

al45tair:
@swift-ci Please test

glessard:
Thanks @al45tair!

PR #180, opened 2024-06-01
Add explicit minimum supported darwin platforms to manifest

roanutil:
It's my understanding that SwiftPM implicitly includes these platform minimums. It would be nice if they're explicitly specified for clarity. For tools that need to parse the manifest, it can make a big difference.

With the minimum tools version of 5.8, visionOS couldn't be included so a 5.9 specific manifest was added.

I based this branch and PR off of dev/merge-1.3.0-back-to-main because that seems to be the current leading branch.

### DIFF
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.8
+// swift-tools-version:5.9
 
 /*
  This source file is part of the Swift System open source project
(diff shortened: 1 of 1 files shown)


glessard:
The package doesn't use any features gated by OS availability, so I'm not clear what this is helpful for?

Lukasa:
For what it's worth, if not done carefully this can also be a semver major for downstream users, as they become forced to add the same annotations to their Package.swift or get unsolveable builds.

glessard:
Can you point us to the url and version involved for Tuist? We should see the problem first-hand, in order to decide if the fix is here or in spm. Thanks!

PR #189, opened 2024-07-17
Ensure that FilePath and SystemString's invariants are enforced when …

stephentyrone:
Decoding these generally takes untrusted input, and corrupted data should be diagnosed with a throw instead of a later crash; the synthesized Decodable conformance has no means to do that for us, so we have to roll our own. It's likely that a more-thorough reworking of these is appropriate, but in the short term we can and should throw an error for the cases that our existing invariants can detect.


stephentyrone:
@swift-ci test

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
```diff
@@ -0,0 +1,127 @@
...
+@testable import SystemPackage
+#else
+@testable import System
+#endif
+
+@available(/*System 0.0.1: macOS 11.0, iOS 14.0, watchOS 7.0, tvOS 14.0*/iOS 8, *)
+final class FilePathDecodableTest: XCTestCase {
+  func testInvalidFilePath() {
+    // _storage is a valid SystemString, but the invariants of FilePath are
+    // violated.
```
milseman, line 22: Which invariant? Is there any way to have a textual representation?

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
```diff
@@ -0,0 +1,127 @@
...
+    XCTAssertThrowsError(try JSONDecoder().decode(
+      FilePath.self,
+      from: Data(input)
+    ))
+  }
+  
+  func testInvalidSystemString() {
+    // _storage is a SystemString whose invariants are violated
+    let input: [UInt8] = [
+      123, 34, 95,115,116,111,114, 97,103,101, 34, 58,123, 34,110,117,108,108,
```
milseman, line 82: Similarly, any way to see what that invariant is?

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
```diff
@@ -0,0 +1,127 @@
...
+      FilePath.self,
+      from: Data(input)
+    ))
+  }
+  
+  func testInvalidExample() {
+    // Another misformed example from Johannes that violates FilePath's
+    // invariants
+    let input: [UInt8] = [
+      123, 34, 95,115,116,111,114, 97,103,101, 34, 58,123, 34,110,117,108,108,
```
milseman, line 100: Similar

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
```diff
@@ -0,0 +1,127 @@
...
+    
+    XCTAssertThrowsError(try JSONDecoder().decode(
+      FilePath.self,
+      from: Data(input)
+    ))
+  }
+  
+  func testEmptyString() {
+    let input: [UInt8] = [
+      123, 34, 95,115,116,111,114, 97,103,101, 34, 58,123, 34,110,117,108,108,
```
milseman, line 117: Same.

milseman (approved):
LGTM, would like to see the erroneous content

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
stephentyrone: The textual representation for these is not very helpful, since the SystemString is basically garbage (hence unnormalized, which is the invariant that fails). We should add additional tests for the other invariants, but these came from examples provided by Johannes in a bug report.

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
stephentyrone: This one has a non-terminating null, I added a comment to explain.

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
stephentyrone: The SystemString encoded here is empty, and hence has no terminating null byte. Added a comment.

stephentyrone:
@swift-ci test

Tests/SystemTests/FilePathTests/FilePathDecodable.swift
weissi: (Just FWIW, when I noticed something was off, I wrote a little (super basic) fuzzer based on just randomness. I then extracted these examples, I don't actually know what they hit.)

PR #191, opened 2024-07-31
Remove private-and-@_aEIC initializers

glessard:
These were just passthroughs, and the combination of `private` with `@_alwaysEmitIntoClient` causes errors with newer compilers.

Resolves rdar://132932883

### DIFF
--- a/Sources/System/Errno.swift
+++ b/Sources/System/Errno.swift
@@ -20,13 +20,10 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   public init(rawValue: CInt) { self.rawValue = rawValue }
 
-  @_alwaysEmitIntoClient
-  private init(_ raw: CInt) { self.init(rawValue: raw) }
-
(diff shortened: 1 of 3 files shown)


glessard:
@swift-ci please test

Sources/System/Errno.swift
```diff
@@ -20,13 +20,12 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   @_alwaysEmitIntoClient
   public init(rawValue: CInt) { self.rawValue = rawValue }
 
-  @_alwaysEmitIntoClient
   private init(_ raw: CInt) { self.init(rawValue: raw) }
```
lorentey, line 23: Is there a reason to keep the private init in this case?

lorentey (approved):

Sources/System/Errno.swift
glessard: I fumbled a conflict resolution. Will fix.

glessard:
@swift-ci please test

PR #205, opened 2024-10-17
[windows] expose pipe’s buffer size parameter

glessard:
The wrapper for Windows's `_pipe()` call should at least expose its memory buffer reservation parameter.
It's perfectly fine to have a default value, so we use the previously hard-coded amount as that default.


lorentey (commented):
👍 

This is just a very small baby step -- a self-confident syscall layer would expose both the buffer size and the flags as _public_ parameters. But it's still a step!

There seems little point to adding defaulted arguments to internal functions -- but at least puts a stake in the ground by diverging the internals.

PR #208, opened 2024-10-29
Draft changes to the io_uring prototype

Catfish-Man:


Sources/CSystem/shims.c
```diff
@@ -1,18 +0,0 @@
-/*
```
Catfish-Man, line 1: This deletion seems probably wrong? Not sure where it came from

Sources/System/AsyncFileDescriptor.swift
```diff
@@ -0,0 +1,166 @@
+@_implementationOnly import CSystem
+
+public struct AsyncFileDescriptor: ~Copyable {
+    @usableFromInline var open: Bool = true
+    @usableFromInline let fileSlot: IORingFileSlot
```
Catfish-Man, line 5: Ideally this should be a borrow of the file slot, but that's hard to express right now

Sources/System/AsyncFileDescriptor.swift
```diff
@@ -0,0 +1,166 @@
...
+            return $0  // bad
+        }
+        let res = await ring.submitAndWait(
+            .openat(
+                atDirectory: atDirectory,
+                path: cstr,
+                mode,
+                options: options,
+                permissions: permissions,
+                intoSlot: fileSlot.borrow()
```
Catfish-Man, line 30: This is a hack, this should also be an actual borrow

Sources/System/AsyncFileDescriptor.swift
```diff
@@ -0,0 +1,166 @@
...
+        if res.result < 0 {
+            throw Errno(rawValue: -res.result)
+        }
+
+        return AsyncFileDescriptor(
+            fileSlot, ring: ring
+        )
+    }
+
+    internal init(_ fileSlot: consuming IORingFileSlot, ring: ManagedIORing) {
```
Catfish-Man, line 41: more "should be a borrow"

Sources/System/AsyncFileDescriptor.swift
```diff
@@ -0,0 +1,166 @@
...
+        self.fileSlot = consume fileSlot
+        self.ring = ring
+    }
+
+    @inlinable @inline(__always)
+    public consuming func close(isolation actor: isolated (any Actor)? = #isolation) async throws {
```
Catfish-Man, line 47: I want to review the heavy use of isolation parameters at some point, as well as inlinability

Sources/System/AsyncFileDescriptor.swift
```diff
@@ -0,0 +1,166 @@
...
+        self.open = false
+    }
+
+    @inlinable @inline(__always)
+    public func read(
+        into buffer: inout UnsafeMutableRawBufferPointer,
+        atAbsoluteOffset offset: UInt64 = UInt64.max,
+        isolation actor: isolated (any Actor)? = #isolation
+    ) async throws -> UInt32 {
+        let file = fileSlot.borrow()
```
Catfish-Man, line 64: more hacks

Sources/System/IORequest.swift
```diff
@@ -0,0 +1,124 @@
+import struct CSystem.io_uring_sqe
+
+public enum IORequest: ~Copyable {
+    case nop  // nothing here
+    case openat(
+        atDirectory: FileDescriptor,
+        path: UnsafePointer<CChar>,
```
Catfish-Man, line 7: `path` should probably be a `FilePath`?

Sources/System/IORing.swift
```diff
@@ -0,0 +1,553 @@
...
+            .baseAddress.unsafelyUnwrapped
+            .advanced(by: sqeIndex)
+
+        ring.userTail = next
+        return sqe
+    }
+    return nil
+}
+
+public struct IORing: @unchecked Sendable, ~Copyable {
```
toffaletti, line 230: Does this need to be Sendable? I found it surprising that SQRing and CQRing are protected by Mutex. If I was writing a single threaded program, I wouldn't want that overhead.

Catfish-Man: yeah, we decided to change that

Sources/System/IORing.swift
```diff
@@ -0,0 +1,657 @@
...
+            || params.features & IORING_FEAT_NODROP == 0
+        {
+            close(ringDescriptor)
+            // TODO: error handling
+            throw IORingError.missingRequiredFeatures
+        }
+
+        if ringDescriptor < 0 {
+            // TODO: error handling
+        }
```
toffaletti, line 232: should this move to line 221 before:

```
if params.features & IORING_FEAT_SINGLE_MMAP == 0
             || params.features & IORING_FEAT_NODROP == 0
         {
             close(ringDescriptor)
```

Sources/System/IORing.swift
```diff
@@ -0,0 +1,657 @@
...
+extension UnsafeMutableRawBufferPointer {
+    func to_iovec() -> iovec {
+        iovec(iov_base: baseAddress, iov_len: count)
+    }
+}
+
+// all pointers in this struct reference kernel-visible memory
+@usableFromInline struct SQRing: ~Copyable {
+    let kernelHead: UnsafePointer<Atomic<UInt32>>
+    let kernelTail: UnsafePointer<Atomic<UInt32>>
```
toffaletti, line 23: should there be padding around these to prevent false sharing?

Catfish-Man: Hmm possibly. Let me look over how they're used, but that's a good point to consider.

Sources/System/IORing.swift
Catfish-Man: I want to redo all the feature detection stuff but I'll take a look at this when I do

Sources/System/IORing.swift
```diff
@@ -0,0 +1,657 @@
...
+            return entry
+        }
+        // TODO: actually block here instead of spinning
+    }
+
+}
+
+//TODO: omitting signal mask for now
+//Tell the kernel that we've submitted requests and/or are waiting for completions
+internal func _enter(
```
hassila, line 116: AFAICT the implementation here for flushing can run into the problem previously encountered here:

https://github.com/apple/swift-nio/blob/140be96c6ba178ce44989b2440e6e116078977e2/Sources/NIOPosix/LinuxUring.swift#L207

See https://github.com/axboe/liburing/issues/309 for background.

Sources/System/IORing.swift
```diff
@@ -0,0 +1,657 @@
...
+    io_uring_sqe
+> {
+    while true {
+        if let entry = _getSubmissionEntry(
+            ring: &ring,
+            submissionQueueEntries: submissionQueueEntries
+        ) {
+            return entry
+        }
+        // TODO: actually block here instead of spinning
```
hassila, line 109: Here you probably want to flush instead, to free up SQE space.

See https://github.com/apple/swift-nio/blob/140be96c6ba178ce44989b2440e6e116078977e2/Sources/NIOPosix/LinuxUring.swift#L185

Sources/System/IOCompletion.swift
```diff
@@ -0,0 +1,56 @@
...
+
+extension IOCompletion {
+    public struct Flags: OptionSet, Hashable, Codable {
+        public let rawValue: UInt32
+
+        public init(rawValue: UInt32) {
+            self.rawValue = rawValue
+        }
+        
+        public static let allocatedBuffer = Flags(rawValue: 1 << 0)
```
hassila, line 15: Would be great to have a link to e.g.

https://github.com/axboe/liburing/blob/92e858ad3f23ab2cdacbb5123ddce4c437452d9f/src/include/liburing/io_uring.h#L446

so one can see where these magic values comes from.

Sources/System/IORing.swift
```diff
@@ -0,0 +1,657 @@
...
+    }
+
+    deinit {
+        munmap(ringPtr, ringSize)
+        munmap(
+            UnsafeMutableRawPointer(submissionQueueEntries.baseAddress!),
+            submissionQueueEntries.count * MemoryLayout<io_uring_sqe>.size
+        )
+        close(ringDescriptor)
+    }
```
Lukasa, line 656: I'm not opposed to the `munmap` calls being here, but I don't think that `close` should be. It's generally not a good idea to clean up resources in deinit, even in the `~Copyable` case. Overwhelmingly the ring will be stored in a reference type of some kind, so the deinit-based cleanup just gets moved out one layer.

I think I'd pretty strongly prefer we put a `close` function here. I'm open to it being `consuming`, though that would also probably be inconvenient too. But generally we shouldn't be getting surprise closes.

This goes double given @CodaFi's note that we want to prevent classes of errors where we shut things down before the resources loaned to the kernel are actually cleaned up. Having an explicit close function gives us a hook on which we can hang that cleanup.

NNNN-swift-system-io-uring.md
```diff
@@ -0,0 +1,471 @@
...
+
+### Blocking
+
+```swift
+let ring = try IORing(queueDepth: 2)
+
+//Make space on the ring for our file (this is optional, but improves performance with repeated use)
+let file = ring.registerFiles(count: 1)[0]
+
+var statInfo = Glibc.stat() // System doesn't have an abstraction for stat() right now
```
rauhul, line 358: FWIW I have this mostly done on a branch

Sources/System/IORing.swift
toffaletti: What are the downsides to doing cleanup in deinit other than not being able to surface any errors? Are they worse than leaking resources?

Catfish-Man:
How concerned should I be about "error: unknown attribute 'unsafe'"? That just seems like a too-old compiler?

Sources/System/IORing.swift
Catfish-Man: oops I totally missed this comment earlier, will take a look!

hassila: Not sure you saw my other comment about flushing either - just sharing some experience from the NIO prototype usage.

jrflat (commented):
Left a few questions but I think it overall looks good! Small nit but maybe we could group the new source files in an IORing folder?

Sources/System/IORing.swift
```diff
@@ -0,0 +1,830 @@
...
+            err = Errno.current
+        }
+        return result
+    }
+
+    if let err {
+        throw err
+    }
+
+    if params.features & IORING_FEAT_SINGLE_MMAP == 0
```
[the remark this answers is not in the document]
Catfish-Man: I think it would be reasonable to do that as a backwards compatibility improvement, I'll see what I can do

Sources/System/IORequest.swift
```diff
@@ -0,0 +1,478 @@
...
+    
+    @inlinable public static func cancel(
+    	_ matchAll: CancellationMatch,
+    	matchingContext: UInt64
+    ) -> IORing.Request {
+        switch matchAll {
+            case .all:
+                .init(core: .cancel(flags: SWIFT_IORING_ASYNC_CANCEL_ALL | SWIFT_IORING_ASYNC_CANCEL_USERDATA, targetContext: matchingContext))
+            case .first:
+                .init(core: .cancel(flags: SWIFT_IORING_ASYNC_CANCEL_ANY | SWIFT_IORING_ASYNC_CANCEL_USERDATA, targetContext: matchingContext))
```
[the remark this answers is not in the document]
jrflat: Or actually I might've misunderstood, would we want to use only `CANCEL_USERDATA` and only `CANCEL_FD` in their respective `.first` cases, since it seems like `IORING_ASYNC_CANCEL_ANY` will cancel any request in the ring regardless of user data or file descriptor match.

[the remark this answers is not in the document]
Catfish-Man: Yeah I think you're right, `_ANY` would be for "no match criteria". Will fix

Catfish-Man:
@swift-ci please test

Catfish-Man:
@swift-ci please test

PR #213, opened 2025-01-23
[cmake] Install libraries in standard directories

Steelskin:
Previously, libs were installed under `lib/swift/${os}`. They should be installed in the default library directory for the relevant target system.
In addition, swiftmodules were installed in the older layout format. This changes to use the standard modern layout format for swiftmodules.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,8 @@ cmake_minimum_required(VERSION 3.16.0)
 project(SwiftSystem
   LANGUAGES C Swift)
 
+include(GNUInstallDirs)
+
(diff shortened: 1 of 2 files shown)


glessard:
@swift-ci please test

glessard:
@swift-ci please test macOS platform

glessard:
Note: will wait to merge until we get a nightly that incorporates the equivalent patch to argument parser.

Steelskin:
> Note: will wait to merge until we get a nightly that incorporates the equivalent patch to argument parser.

FYI, this landed in apple/swift-argument-parser#685

PR #215, opened 2025-02-05
OpenBSD support.

3405691582:
Two commits:

* straightforward platform conditional to typealias `system_DIRPtr` properly (it would be ideal if the `#else` case was an `#error` to guide new porters appropriately, but that's another show).
* `ManagedBuffer.capacity` is unavailable on OpenBSD, so must be handled specially. Thankfully the workaround is straightforward here, I think.

### DIFF
--- a/Sources/System/Internals/Syscalls.swift
+++ b/Sources/System/Internals/Syscalls.swift
@@ -191,7 +191,7 @@ internal func system_confstr(
 internal let SYSTEM_AT_REMOVE_DIR = AT_REMOVEDIR
 internal let SYSTEM_DT_DIR = DT_DIR
 internal typealias system_dirent = dirent
-#if os(Linux) || os(Android) || os(FreeBSD)
+#if os(Linux) || os(Android) || os(FreeBSD) || os(OpenBSD)
 internal typealias system_DIRPtr = OpaquePointer
 #else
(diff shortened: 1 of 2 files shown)


glessard:
@swift-ci please test

glessard (approved):

Sources/System/Internals/Syscalls.swift
```diff
@@ -191,7 +191,7 @@ internal func system_confstr(
 internal let SYSTEM_AT_REMOVE_DIR = AT_REMOVEDIR
 internal let SYSTEM_DT_DIR = DT_DIR
 internal typealias system_dirent = dirent
-#if os(Linux) || os(Android) || os(FreeBSD)
+#if os(Linux) || os(Android) || os(FreeBSD) || os(OpenBSD)
 internal typealias system_DIRPtr = OpaquePointer
 #else
 internal typealias system_DIRPtr = UnsafeMutablePointer<DIR>
```
jakepetroules, line 197: What exactly is "else" here? I'm guessing only Darwin falls here in practice. Would this be more stable to accommodate new platforms if written as if-Darwin-else?

jakepetroules: Hmm, seems like it, given the if !Windows above. Anyways, understandable to leave it for now.

PR #187, opened 2024-07-04
Import new Android overlay

finagolfin:
This new overlay was recently added to trunk in swiftlang/swift#72161 and swiftlang/swift#72634, then to 6.0 in swiftlang/swift#74758. I've been building it with this patch on my daily Android CI since the trunk addition, finagolfin/swift-android-sdk#151, and natively on Android too.

@glessard, should be an easy review.

PR #225, opened 2025-03-13
Add Android imports to tests

marcprux:
This is needed for constants like `O_RDWR`, completing the work in https://github.com/apple/swift-system/pull/187 and eliminating the need for @finagolfin's [swift-system-tests.patch](https://github.com/finagolfin/swift-android-sdk/blob/main/swift-system-tests.patch)

### DIFF
--- a/Tests/SystemTests/FileOperationsTest.swift
+++ b/Tests/SystemTests/FileOperationsTest.swift
@@ -14,6 +14,9 @@ import XCTest
 #else
 @testable import System
 #endif
+#if canImport(Android)
+import Android
+#endif
 
 @available(/*System 0.0.1: macOS 11.0, iOS 14.0, watchOS 7.0, tvOS 14.0*/iOS 8, *)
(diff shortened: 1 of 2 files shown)


finagolfin:
The reason I haven't submitted that patch is that I only need it when cross-compiling, ie I don't need it when natively compiling on Android. There are [some weird import issues I've been seeing since last summer, possibly related to the Foundation rewrite](https://github.com/apple/swift-certificates/pull/183#discussion_r1730366251). Since this is one of them, I've been holding off on submitting it, as I want to investigate the root cause at some point.

However, if you want to just fix this for cross-compilation now, nothing against that: go for it.

jakepetroules:
@swift-ci test

jakepetroules (approved):

finagolfin:
@jakepetroules, would you merge too? We don't have commit access on this repo.

glessard (approved):

PR #226, opened 2025-04-18
Transparently add the \\?\ prefix to Win32 calls for extended length path handling

jakepetroules:
On Windows, there is a built-in maximum path limitation of 260 characters under most conditions. This can be extended to 32767 characters under either of the following two conditions:

- Adding the longPathAware attribute to the executable's manifest AND enabling the LongPathsEnabled system-wide registry key or group policy.
- Ensuring fully qualified paths passed to Win32 APIs are prefixed with \\?\

Unfortunately, the former is not realistic for the Swift ecosystem, since it requires developers to have awareness of this specific Windows limitation, AND set longPathAware in their apps' manifest AND expect end users of those apps to change their system configuration.

Instead, this patch transparently prefixes all eligible paths in calls to Win32 APIs with the \\?\ prefix to allow them to work with paths longer than 260 characters without requiring the caller of System to manually prefix the paths.

See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation for more info.

### DIFF
--- a/Sources/System/FilePath/FilePathWindows.swift
+++ b/Sources/System/FilePath/FilePathWindows.swift
@@ -461,3 +461,78 @@ extension SystemString {
     return lexer.current
   }
 }
+
+#if os(Windows)
+import WinSDK
+
+// FIXME: Rather than canonicalizing the path at every call site to a Win32 API,
+// we should consider always storing absolute paths with the \\?\ prefix applied,
+// for better performance.
+extension UnsafePointer where Pointee == CInterop.PlatformChar {
+  /// Invokes `body` with a resolved and potentially `\\?\`-prefixed version of the pointee,
+  /// to ensure long paths greater than MAX_PATH (260) characters are handled correctly.
+  ///
+  /// - seealso: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+  internal func withCanonicalPathRepresentation<Result>(_ body: (Self) throws -> Result) throws -> Result {
+    // 1. Normalize the path first.
+    // Contrary to the documentation, this works on long paths independently
+    // of the registry or process setting to enable long paths (but it will also
+    // not add the \\?\ prefix required by other functions under these conditions).
+    let dwLength: DWORD = GetFullPathNameW(self, 0, nil, nil)
+    return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) { pwszFullPath in
+      guard (1..<dwLength).contains(GetFullPathNameW(self, DWORD(pwszFullPath.count), pwszFullPath.baseAddress, nil)) else {
+        throw Errno(rawValue: _mapWindowsErrorToErrno(GetLastError()))
+      }
+
+      // 1.5 Leave \\.\ prefixed paths alone since device paths are already an exact representation and PathCchCanonicalizeEx will mangle these.
+      if let base = pwszFullPath.baseAddress,
+        base[0] == UInt8(ascii: "\\"),
+        base[1] == UInt8(ascii: "\\"),
+        base[2] == UInt8(ascii: "."),
+        base[3] == UInt8(ascii: "\\") {
+        return try body(base)
+      }
+
+      // 2. Canonicalize the path.
+      // This will add the \\?\ prefix if needed based on the path's length.
+      var pwszCanonicalPath: LPWSTR?
+      let flags: ULONG = numericCast(PATHCCH_ALLOW_LONG_PATHS.rawValue)
+      let result = PathAllocCanonicalize(pwszFullPath.baseAddress, flags, &pwszCanonicalPath)
+      if let pwszCanonicalPath {
+          defer { LocalFree(pwszCanonicalPath) }
+          if result == S_OK {
+            // 3. Perform the operation on the normalized path.
+            return try body(pwszCanonicalPath)
+          }
+      }
+      throw Errno(rawValue: _mapWindowsErrorToErrno(WIN32_FROM_HRESULT(result)))
+    }
+  }
+}
+
+@inline(__always)
+fileprivate func HRESULT_CODE(_ hr: HRESULT) -> DWORD {
+    DWORD(hr) & 0xffff
+}
+
+@inline(__always)
+fileprivate func HRESULT_FACILITY(_ hr: HRESULT) -> DWORD {
+    DWORD(hr << 16) & 0x1fff
+}
+
+@inline(__always)
+fileprivate func SUCCEEDED(_ hr: HRESULT) -> Bool {
+    hr >= 0
+}
+
+// This is a non-standard extension to the Windows SDK that allows us to convert
+// an HRESULT to a Win32 error code.
+@inline(__always)
+fileprivate func WIN32_FROM_HRESULT(_ hr: HRESULT) -> DWORD {
+    if SUCCEEDED(hr) { return DWORD(ERROR_SUCCESS) }
+    if HRESULT_FACILITY(hr) == FACILITY_WIN32 {
+        return HRESULT_CODE(hr)
+    }
+    return DWORD(hr)
+}
+#endif
--- a/Sources/System/Internals/WindowsSyscallAdapters.swift
+++ b/Sources/System/Internals/WindowsSyscallAdapters.swift
@@ -35,17 +35,17 @@ internal func open(
     bInheritHandle: decodedFlags.bInheritHandle
   )
 
-  let hFile = CreateFileW(path,
-                          decodedFlags.dwDesiredAccess,
-                          DWORD(FILE_SHARE_DELETE
-                                  | FILE_SHARE_READ
-                                  | FILE_SHARE_WRITE),
-                          &saAttrs,
-                          decodedFlags.dwCreationDisposition,
-                          decodedFlags.dwFlagsAndAttributes,
-                          nil)
-
-  if hFile == INVALID_HANDLE_VALUE {
+  guard let hFile = try? path.withCanonicalPathRepresentation({ path in
+    CreateFileW(path,
+                decodedFlags.dwDesiredAccess,
+                DWORD(FILE_SHARE_DELETE
+                      | FILE_SHARE_READ
+                      | FILE_SHARE_WRITE),
+                &saAttrs,
+                decodedFlags.dwCreationDisposition,
+                decodedFlags.dwFlagsAndAttributes,
+                nil)
+  }), hFile != INVALID_HANDLE_VALUE else {
     ucrt._set_errno(_mapWindowsErrorToErrno(GetLastError()))
     return -1
   }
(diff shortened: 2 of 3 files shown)


jakepetroules:
I anticipate there _might_ be some contention about this patch given the general spirit of SwiftSystem not incorporating higher level cross platform translations/behaviors (which should go in Foundation) but wanted to send this up to prompt some discussion of this idea.

I'd like to point out that we already convert UTF-8 strings to UTF-16 _and_ normalize slashes in file paths, so I feel that adding the extended length prefix is still within the spirit of the level of platform translations that SwiftSystem should be doing.

Without doing this, a lot of programs will simply fail to access files whose paths are over the length limit; it's very unrealistic to assume that developers (even those relatively familiar with Windows) know about \\?\ or will properly propagate it everywhere it's needed.

Do other folks agree?

Sources/System/FilePath/FilePathWindows.swift
```diff
@@ -461,3 +461,29 @@ extension SystemString {
...
+  ///
+  /// In all other cases, invokes the body with the original pointer.
+  ///
+  /// - precondition: The pointee should already have normalized path separators to `\`.
+  /// - seealso: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+  internal func withExtendedLengthPrefixIfPossible<T>(_ body: (Self) -> T) -> T {
+    if ((self[0] >= UInt8(ascii: "A") && self[0] <= UInt8(ascii: "Z")) || (self[0] >= UInt8(ascii: "a") && self[0] <= UInt8(ascii: "z")))
+        && self[1] == UInt8(ascii: ":")
+        && self[2] == UInt8(ascii: "\\") {
+      return (extendedLengthPathPrefix + Array(UnsafeBufferPointer(start: self, count: system_platform_strlen(self) + 1))).withUnsafeBufferPointer { ptr in
```
compnerd, line 483: The way that I did this in Foundation was:

```swift
let dwLength = GetFullPathNameW(pwszPath, 0, nil, nil)
return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
  guard GetFullPathNameW(pwszPath, DWORD($0.count), $0.baseAddress, nil) > 0 else {
    throw CocoaError.errorWithFilePath(self, win32: GetLastError(), reading: true)
  }

  // 2. Perform the operation on the normalized path.
  return try body($0.baseAddress!)
}
```

This avoids us having to account for the various path normalization behaviours of Windows.

glessard:
This approach doesn't seem unreasonable to me. It's about setting up the syscalls for success, just as we've gone to significant lengths not to clobber the errno state prematurely.

jakepetroules:
@swift-ci test

Sources/System/FilePath/FilePathWindows.swift
jakepetroules: Based on some offline discussion... I adopted your approach of using GetFullPathNameW, since (contrary to the documentation) it's capable of handling paths >260 chars regardless of the registry or process long paths enablement. However, we also need to call PathCchCanonicalizeEx to add the \\?\ prefix, since GetFullPathNameW will not do that, and the remaining Win32 path functions are expected to need it.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

Sources/System/FilePath/FilePathWindows.swift
```diff
@@ -461,3 +461,46 @@ extension SystemString {
...
+  ///
+  /// - seealso: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+  internal func withCanonicalPathRepresentation<Result>(_ body: (Self) throws -> Result) throws -> Result {
+    // 1. Normalize the path first.
+    // Contrary to the documentation, this works on long paths independently
+    // of the registry or process setting to enable long paths (but it will also
+    // not add the \\?\ prefix required by other functions under these conditions).
+    let dwLength: DWORD = GetFullPathNameW(self, 0, nil, nil)
+    return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) { fullPath in
+      guard GetFullPathNameW(self, DWORD(fullPath.count), fullPath.baseAddress, nil) > 0 else {
```
lorentey, line 480: I believe it needs to be verified that the buffer was still large enough to hold the result here, as the current working directory can change between `GetFullPathNameW` invocations.

Sources/System/FilePath/FilePathWindows.swift
```diff
@@ -461,3 +461,46 @@ extension SystemString {
...
+
+      // 1.5 Leave \\.\ prefixed paths alone since device paths are already an exact representation and PathCchCanonicalizeEx will mangle these.
+      if let base = fullPath.baseAddress, base[0] == UInt8(ascii: "\\"), base[1] == UInt8(ascii: "\\"), base[2] == UInt8(ascii: "."), base[3] == UInt8(ascii: "\\") {
+        return try body(base)
+      }
+
+      // 2. Canonicalize the path.
+      // This will add the \\?\ prefix if needed based on the path's length.
+      let capacity = Int16.max
+      return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: numericCast(capacity)) { outBuffer in
```
lorentey, line 492: This is allocating 32767 bytes, and I expect that will be executed as a heap allocation, repeated every time this is called. Usually swift-system tries to avoid doing implicit allocations like that, as clients sometimes get infuriated when they find out that a low-level syscall wrapper library engages in those.

We may have already lost those clients on Windows; I worry that this may make it harder to try winning them back.

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -35,15 +35,23 @@ internal func open(
...
+                  DWORD(FILE_SHARE_DELETE
+                        | FILE_SHARE_READ
+                        | FILE_SHARE_WRITE),
+                  &saAttrs,
+                  decodedFlags.dwCreationDisposition,
+                  decodedFlags.dwFlagsAndAttributes,
+                  nil)
+    })
+  } catch {
+    ucrt._set_errno(_mapWindowsErrorToErrno(GetLastError()))
```
lorentey, line 52: I do not think it is right to set the errno here. (Or anywhere in swift-system.) 

These _internal_ functions should report system errors (or HRESULTs) by throwing the actual error. (As a typed error, if our toolchain version is already high enough.)

Sources/System/FilePath/FilePathWindows.swift
```diff
@@ -461,3 +461,46 @@ extension SystemString {
...
+
+#if os(Windows)
+import WinSDK
+
+extension UnsafePointer where Pointee == CInterop.PlatformChar {
+  /// Invokes `body` with a resolved and potentially `\\?\`-prefixed version of the pointee,
+  /// to ensure long paths greater than 260 characters are handled correctly.
+  ///
+  /// - seealso: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+  internal func withCanonicalPathRepresentation<Result>(_ body: (Self) throws -> Result) throws -> Result {
```
lorentey, line 473: Should `FilePath` offer this as a public operation?

Sources/System/FilePath/FilePathWindows.swift
jakepetroules: Right, it should be `== dwLength - 1` rather than `> 0`

Sources/System/FilePath/FilePathWindows.swift
jakepetroules: In terms of any further high level business logic feedback, we should probably keep that on swiftlang/swift-foundation#1257 since I'm effectively copying that modified implementation over here to swift-system and swift-build as well.

Sources/System/Internals/WindowsSyscallAdapters.swift
jakepetroules: These functions are already calling _set_errno now, so I think that's something that should come in a future enhancement, not entangled with this change. (plus they are non-throwing)

Sources/System/FilePath/FilePathWindows.swift
jakepetroules: I changed this to use PathAllocCanonicalize, which has Windows allocate the buffer and can theoretically be smaller, but it's always going to be a heap allocation.

That said, Win32 APIs are not syscalls; they are higher level than the Unix "equivalents". Windows does not have a stable syscall interface and thus it's undocumented. So I think we can get away with some heap allocations where necessary to use the APIs correctly, since many Win32 functions are already doing so internally.

Sources/System/FilePath/FilePathWindows.swift
jakepetroules: Fixed!

Sources/System/FilePath/FilePathWindows.swift
jakepetroules: It's tempting, but I'd rather leave that to a future enhancement, I'm not sure yet that this should be a public API or what the right shape would be for it, since we'd have to commit to supporting it if we were to make that change.

jakepetroules:
@swift-ci test

Sources/System/FilePath/FilePathWindows.swift
al45tair: I _think_ the length of the result of the `Path*Canonicalize*` functions is always less than or equal to the length of the input string plus one (plus two if you've got a trailing `NUL`), so we don't need to allocate 32KB here.

I also wonder whether we could have a canonicalize-in-place function that mutates the underlying `SystemString` as a more efficient option.

jakepetroules:
@swift-ci test

Sources/System/Internals/WindowsSyscallAdapters.swift
compnerd: why not change this to a

```swift
guard let hFile = try? ... else {
  ucrt._set_errno(...)
  return -1
}
```

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -242,7 +258,12 @@ internal func mkdir(
     bInheritHandle: false
   )
 
-  if !CreateDirectoryW(path, &saAttrs) {
+  do {
+    if try !path.withCanonicalPathRepresentation({ path in CreateDirectoryW(path, &saAttrs) }) {
```
compnerd, line 262: It might be nicer as `guard try? .... else { }` I think.

jakepetroules: Done

jakepetroules:
@swift-ci test

Sources/System/Internals/WindowsSyscallAdapters.swift
jakepetroules: Done.

glessard (approved):
Looks good, but it would be nice to have a test for the addition in FilePathWindows.

PR #230, opened 2025-05-22
Fix Windows path parsing when the package is built in release configuration

owenv:
forceWindowsPaths can be used to force Windows path parsing both on and off. Instead of defaulting it to false in release builds, we should not express any preference. This fixes path parsing on Windows in optimized builds


owenv:
@swift-ci test

jakepetroules (approved):

glessard (approved):

PR #239, opened 2025-07-25
Fix WASI build (dirent, umask, errno and open flag tests)

jrflat:
Builds successfully with `swift-DEVELOPMENT-SNAPSHOT-2025-07-23-a_wasm`. Tests also build but do not run (same symptom as https://forums.swift.org/t/swiftpm-fails-to-build-swift-testing-targeting-wasm/81085).

- `_getConst_DT_DIR()` for `DT_DIR`
- Modified `_system_dirent` struct that can be imported to Swift (WASI's `dirent` has variable length `char d_name[]` which can't be imported, so this uses `NAME_MAX + 1` instead)
- `#if !os(WASI)` unavailable functions like `umask()` and `pipe()`, as well as unavailable `Errno` and open flag constants that are used directly in tests

### DIFF
--- a/Sources/CSystem/include/CSystemWASI.h
+++ b/Sources/CSystem/include/CSystemWASI.h
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift System open source project
 
- Copyright (c) 2024 Apple Inc. and the Swift System project authors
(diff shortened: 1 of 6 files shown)


jrflat:
@swift-ci please test

PR #243, opened 2025-07-28
[CI] expand github actions

glessard:
Test more configurations


glessard:
The commit was included in https://github.com/apple/swift-system/pull/247

Issue #240, opened 2025-07-26
`'linux/io_uring.h'` file not found


zaneenders:
Hi, 

`main` fails to build on Linux due to missing `linux/io_uring.h`.

```
➜  swift-system git:(main) swift build -c release --swift-sdk aarch64-swift-linux-musl

Building for production...
In file included from /home/zane/.scribe/Projects/swift-system/Sources/CSystem/shims.c:12:
In file included from /home/zane/.scribe/Projects/swift-system/Sources/CSystem/include/CSystemLinux.h:24:
/home/zane/.scribe/Projects/swift-system/Sources/CSystem/include/io_uring.h:6:10: fatal error: 'linux/io_uring.h' file not found
    6 | #include <linux/io_uring.h>
      |          ^~~~~~~~~~~~~~~~~~
1 error generated.
[1/3] Write swift-version--369F648853B632EA.txt


➜  swift-system git:(main) ls /usr/include/linux/io_uring.h                           
/usr/include/linux/io_uring.h

➜  swift-system git:(main) uname -r
6.14.8-400.asahi.fc42.aarch64+16k
```

zaneenders:
Version `1.5.0` builds.

Lukasa:
This also affects the static linux SDK

Issue #237, opened 2025-07-25
1.6.0 release is not buildable with Swift 6.2 in release mode


vsarunas:
@Catfish-Man  - debug version compiles but not in release mode, after merge of https://github.com/apple/swift-system/pull/208
```
error: compile command failed due to signal 6 (use -v to see invocation)
SIL memory lifetime failure in @$s13SystemPackage6IORingV10queueDepth5flagsACs6UInt32V_AC10SetupFlagsVtAA5ErrnoVYKcfC: memory is not initialized, but should be
memory location:   %195 = struct_element_addr %3 : $*IORing, #IORing.submissionRing // users: %214, %196
at instruction:   %196 = load [take] %195 : $*SQRing              // user: %197

in function:
// IORing.init(queueDepth:flags:)
// Isolation: unspecified
sil [ossa] @$s13SystemPackage6IORingV10queueDepth5flagsACs6UInt32V_AC10SetupFlagsVtAA5ErrnoVYKcfC : $@convention(method) (UInt32, IORing.SetupFlags, @thin IORing.Type) -> (@owned IORing, @error Errno) {
[global: read,write,copy,destroy,allocate,deinit_barrier]
// %0 "queueDepth"                                // users: %20, %4
// %1 "flags"                                     // users: %20, %5
// %2 "$metatype"
bb0(%0 : $UInt32, %1 : $IORing.SetupFlags, %2 : $@thin IORing.Type):
  %3 = alloc_stack [lexical] [var_decl] $IORing, var, name "self", type $IORing // users: %45, %47, %49, %51, %53, %55, %57, %65, %71, %261, %385, %387, %7, %11, %363, %195, %392, %393, %399, %389
  debug_value %0, let, name "queueDepth", argno 1 // id: %4
  debug_value %1, let, name "flags", argno 2      // id: %5
  debug_value undef : $Errno, var, name "$error", argno 3 // id: %6
  %7 = struct_element_addr %3, #IORing.submissionPolling // user: %10
  %8 = integer_literal $Builtin.Int1, 0           // user: %9
  %9 = struct $Bool (%8)                          // user: %10
  store %9 to [trivial] %7                        // id: %10
  %11 = struct_element_addr %3, #IORing.features  // users: %44, %17
  %12 = integer_literal $Builtin.Int32, 0         // user: %13
  %13 = struct $UInt32 (%12)                      // users: %213, %16, %15, %14
  debug_value %13, let, name "rawValue", argno 1  // id: %14
[...]

PR #246, opened 2025-07-29
[CI] Add workflows for release and static SDK builds

jrflat:
Adds Github workflows to build System in release mode and with the static Linux SDK. These workflows should prevent issues like #237 and #240 in the future.

The release build workflow is pretty simple and uses the `swift_package_test.yml@main` workflow from the main swiftlang repo. Installing and building with the static SDK is more involved since we may need to download a new toolchain to match the snapshot tag from the most recent static SDK. For now, only 6.2 static SDK is tested, but we could take a version parameter in the future. I tested it on a main image and it downloads and tests with the appropriate 6.2 toolchain to match the most recent 6.2 static SDK. Happy to learn if there's a more idiomatic way of testing static SDK builds.

Resolves rdar://156942820

### DIFF
--- a/.github/workflows/scripts/install-and-build-with-6.2-static-sdk.sh
+++ b/.github/workflows/scripts/install-and-build-with-6.2-static-sdk.sh
@@ -23,15 +23,18 @@ detect_os_info() {
         fatal "Cannot detect OS: /etc/os-release not found"
     fi
 
-    local os_id=$(grep '^ID=' /etc/os-release | cut -d'=' -f2 | tr -d '"' | tr '[:upper:]' '[:lower:]')
-    local version_id=$(grep '^VERSION_ID=' /etc/os-release | cut -d'=' -f2 | tr -d '"')
+    local os_id
+    os_id=$(grep '^ID=' /etc/os-release | cut -d'=' -f2 | tr -d '"' | tr '[:upper:]' '[:lower:]')
+    local version_id
+    version_id=$(grep '^VERSION_ID=' /etc/os-release | cut -d'=' -f2 | tr -d '"')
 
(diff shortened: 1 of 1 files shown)


jrflat:
Static SDK build failed as expected. However, the release builds did not fail. I wonder if we need to `rm -rf .build/` before building

jrflat:
Currently based on `release/1.6.0`:
- Build using static Linux SDK is failing as expected
- Linux release builds using Swift 6.2 and main toolchains are failing as expected
- Focal builds are failing as expected (due to older kernel version)

I'll rebase back to `main` and verify these checks all pass again.

Screenshot 2025-07-29 at 4 21 45 PM

glessard:
It adds up fast, and we haven't added testing on ubuntu 24.04 yet!
I wonder if we could get a number of these to run once nightly using the swift-ci infrastructure instead.

rauhul:
I feel like these tests should be added to the upstream GitHub workflow so all repos can get coverage

jrflat:
> It adds up fast, and we haven't added testing on ubuntu 24.04 yet! I wonder if we could get a number of these to run once nightly using the swift-ci infrastructure instead.

Yeah, I'm wondering if it's better to just test a couple configurations that are likely to catch issues, like (Ubuntu 20.04, 24.04) x (Swift 6.2, main). I think the static SDK test, which only tests a single configuration (Ubuntu 24.04 x Swift 6.2), should realistically catch almost all static SDK-specific build failures.

jrflat:
Linux static SDK and Wasm SDK build jobs were added in the upstream `swiftlang/github-workflows` repo in https://github.com/swiftlang/github-workflows/pull/142

I'm going to close this PR and open a new PR that enables these jobs for swift-system

PR #252, opened 2025-07-31
Support testing in release mode

jrflat:
Some tests used internal types/functions that are guarded by `#if ENABLE_MOCKING` (defined in debug builds), but their code wasn't guarded by `#if ENABLE_MOCKING`. This made the tests fail to build in release mode.

This PR guards the respective tests/types in `#if ENABLE_MOCKING` so we can run `swift test -c release --enable-testable-imports` as suggested in https://github.com/swiftlang/github-workflows/pull/134.


jrflat:
@swift-ci please test

PR #254, opened 2025-08-03
Fix typedef errors on platforms with _ASM_GENERIC_INT headers

jrflat:
System fails to build on Amazon Linux 2 with the error:

```
/Swift/swift-system/Sources/CSystem/include/io_uring.h:141:18: error: typedef redefinition with different types ('uint64_t' (aka 'unsigned long') vs 'unsigned long long')
  141 | typedef uint64_t __u64;
      |                  ^
/usr/include/asm-generic/int-ll64.h:31:42: note: previous definition is here
   31 | __extension__ typedef unsigned long long __u64;
      |                                          ^
```

We can guard the `__u64`, `__u32`, etc. typedefs with `#if !defined` for the various `_ASM_GENERIC_INT` headers where these types are already defined.

Verified System builds successfully on Amazon Linux 2 with this change.


jrflat:
@swift-ci please test

glessard (approved):

PR #270, opened 2025-11-04
Fix Stat build on Android

jrflat:
On Android on certain architectures, `struct stat` defines properties
```
unsigned long long st_dev;
unsigned long long st_ino;
unsigned int st_mode;
```
instead of
```
dev_t st_dev;
ino_t st_ino;
mode_t st_mode;
```

I think we should keep the `dev_t`, `ino_t`, and `mode_t` types as the backing for `CInterop.DeviceID`, `CInterop.Inode`, and `CInterop.Mode`, so this PR just performs a `numericCast(_:)` in the respective `Stat` getter and setters, which is compiled out in cases where the types match.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -46,6 +46,10 @@ jobs:
           {"os_version": "focal", "swift_version": "nightly-main"},
         ]
       enable_android_sdk_build: true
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

glessard (approved):

PR #273, opened 2025-12-08
[IO_URING] Add support for `pollAdd` operation

FranzBusch:
## Motivation

To observe events on file descriptors IO_URING supports the `pollAdd` operation. This is useful when you want to observe a file descriptor becoming ready to read or write

## Modifications

This PR adds a new `IORing.Request.PollEvents` option set to model the poll masks. Furthermore, it adds a new `static func pollAdd` to the `IORing.Request`.

## Result

We can now use IO_URING to poll for events on file descriptors.

### DIFF
--- a/Sources/System/IORing/IORequest.swift
+++ b/Sources/System/IORing/IORequest.swift
@@ -256,7 +256,7 @@ extension IORing.Request {
         isMultiShot: Bool = false,
         context: UInt64 = 0
     ) -> IORing.Request {
-        .init(core: .pollAdd(file: file, pollEvents: pollEvents, context: context))
+        .init(core: .pollAdd(file: file, pollEvents: pollEvents, isMultiShot: isMultiShot, context: context))
     }
     
     @inlinable public static func read(
--- a/Sources/System/IORing/IORequest.swift
+++ b/Sources/System/IORing/IORequest.swift
@@ -196,8 +196,8 @@ extension IORing.Request {
     /// Adds a poll operation to monitor a file descriptor for specific I/O events.
     ///
     /// This method creates an io_uring poll operation that monitors the specified file descriptor
-    /// for I/O readiness events. The operation completes when any of the requested events become
-    /// active on the file descriptor, such as data becoming available for reading or the descriptor
+    /// for I/O readiness events. The operation completes when any of the requested events
+    /// occur on the file descriptor, such as data becoming available for reading or the descriptor
     /// becoming ready for writing.
     ///
     /// Poll operations are useful for implementing efficient I/O multiplexing, allowing you to
--- a/Sources/System/IORing/IORequest.swift
+++ b/Sources/System/IORing/IORequest.swift
@@ -609,7 +609,7 @@ extension IORing.Request {
             if isMultiShot {
                 request.rawValue.len = Self.SWIFT_IORING_POLL_ADD_MULTI
             }
-            request.rawValue.poll32_events = pollEvents.rawValue
+            request.pollEvents = pollEvents
         }
 
         return request
--- a/Sources/System/IORing/IORequest.swift
+++ b/Sources/System/IORing/IORequest.swift
@@ -23,6 +23,12 @@ internal enum IORequestCore {
         intoSlot: IORing.RegisteredFile,
         context: UInt64 = 0
     )
+    case pollAdd(
+        file: FileDescriptor,
+        pollEvents: IORing.Request.PollEvents,
+        isMultiShot: Bool = true,
+        context: UInt64 = 0
+    )
     case read(
         file: FileDescriptor,
         buffer: IORing.RegisteredBuffer,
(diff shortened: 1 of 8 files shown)


Sources/System/IORing/PollEvents.swift
```diff
@@ -0,0 +1,59 @@
...
+            self.rawValue = rawValue
+        }
+
+        /// An event indicating data is available for reading.
+        ///
+        /// This event becomes active when data arrives on the file descriptor and can be read
+        /// without blocking. For sockets, this includes when a new connection is available on
+        /// a listening socket. Corresponds to the Posix `POLLIN` event flag.
+        @inlinable
+        public static var pollin: PollEvents { PollEvents(rawValue: 0x0001) }
```
Catfish-Man: Should this and pollout be camel case?

Sources/System/IORing/IORequest.swift
```diff
@@ -187,6 +193,72 @@ extension IORing.Request {
...
+    /// - Returns: An I/O ring request that monitors the file descriptor for the specified events.
+    ///
+    /// ## See Also
+    ///
+    /// - ``PollEvents``: The events that can be monitored.
+    /// - ``IORing/Request/cancel(_:matching:)``: Cancelling poll operations.
+    @inlinable public static func pollAdd(
+        _ file: FileDescriptor,
+        pollEvents: PollEvents,
+        isMultiShot: Bool = true,
```
Catfish-Man: Do we want multishot to be true by default? Past experience with NSFileHandle's readabilityHandler makes me think one-shot monitoring is a bit easier to handle and meets a lot of clients needs.

Catfish-Man (dismissed):
I had a few comments but I don't think any of them are necessarily blockers

Sources/System/IORing/PollEvents.swift
```diff
@@ -0,0 +1,59 @@
...
+    ///     socketFD,
+    ///     pollEvents: .pollin,
+    ///     isMultiShot: true
+    /// )
+    /// ```
+    public struct PollEvents: OptionSet, Hashable, Codable {
+        public var rawValue: UInt32
+
+        @inlinable
+        public init(rawValue: UInt32) {
```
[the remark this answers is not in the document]
Catfish-Man: Ah, right

FranzBusch:
@swift-ci please test

glessard (commented):
I'd be inclined to approve this addition if that were our process.

FranzBusch:
@swift-ci please test

glessard:
@swift-ci please test

Sources/System/IORing/PollEvents.swift
```diff
@@ -0,0 +1,112 @@
...
+            self.rawValue = event.rawValue
+        }
+
+        @usableFromInline
+        enum Event: UInt32, RawRepresentable, Hashable, CaseIterable {
+            case pollIn = 0x0001
+            case pollOut = 0x0004
+            case pollErr = 0x0008
+            case pollHup = 0x0010
+            case pollNval = 0x0020
```
glessard, line 54: Added cases for POLLHUP and POLLNVAL. These names are inscrutable in any situation. Should they be pollHup and pollNval, or pollHangup and pollInvalid, or hangUp and invalidDescriptor. For now I continued along with the earlier work in this PR and used what I consider to be the bad names.

Missing are POLLPRI and POLLRDHUP, which would be relevant for sockets.

Better names would be `readable`, `writable`, `error`, `hangUp`, `invalidDescriptor` (+ `priorityData`, and `peerClosed`)

glessard:
@swift-ci please test

Sources/System/IORing/IORequest.swift
```diff
@@ -187,6 +193,72 @@ extension IORing.Request {
...
+    ///   - context: An application-specific value passed through to the completion event,
+    ///     allowing you to identify which operation completed. Defaults to `0`.
+    ///
+    /// - Returns: An I/O ring request that monitors the file descriptor for the specified events.
+    ///
+    /// ## See Also
+    ///
+    /// - ``PollEvents``: The events that can be monitored.
+    /// - ``IORing/Request/cancel(_:matching:)``: Cancelling poll operations.
+    @inlinable public static func pollAdd(
```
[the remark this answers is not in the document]
glessard: `pollAdd` is somewhat greppable with io_uring docs, but not directly since the names are usually something like `io_uring_prep_poll_add` or `IORING_OP_POLL_ADD`. This is the name for the request, so maybe just `poll` would be better: `poll(fd, events: PollEvents, isMultiShot: Bool, context: UInt64)`.

A future addition would be removal of events, and the constant used for that in the C library is `IORING_POLL_UPDATE_EVENTS`. I don't think we would want `pollUpdate` or `pollRemove`, maybe just `update`.

glessard:
Public symbols added in this PR:
```swift
struct IORing.Request.PollEvents: OptionSet, Hashable, Codable, CaseIterable {
  init(rawValue: UInt32)
  static var allCases: [PollEvents]
  static var pollIn   // 0x0001
  static var pollOut  // 0x0004
  static var pollErr  // 0x0008
  static var pollHup  // 0x0010
  static var pollNval // 0x0020
}

extension IORing.Request {
  static func pollAdd(_:pollEvents:isMultiShot:context:)
}
```
Let's figure out whether these are the names we want.

I think the presence of `poll` on everything is a distraction. `PollEvents` is fine, but the 5 flags could have better names. I'm also not convinced by `pollAdd` as a case name on `Request`.

I believe having some accessors on `Completion` to obtain information would be useful. Something coarse would be a `Result<PollEvents, Errno>`-returning accessor. Any `Completion` improvements can be separate, but we can note that the usage examples (tests) don't look ergonomic.

FranzBusch:
> I think the presence of `poll` on everything is a distraction. `PollEvents` is fine, but the 5 flags could have better names.

I think this is a good suggestion and just renaming them to `.in`, `.out`, `.error`, `.hub`, etc would be better. I am not sure if we should give them completely new names such as `readable` or `writable`. Users of this API are most likely looking at man pages for liburing and there is a lot of value sticking close to that documentation.

> I'm also not convinced by `pollAdd` as a case name on `Request`.

Similarly here I personally think that sticking as close to the libuiring and ops code naming is beneficial when it comes to discovery.

glessard:
The pattern we've used elsewhere it to cover the c-style name as a deprecated symbol, with a message indicating how it has been renamed. Would that be reasonable?

FranzBusch:
> The pattern we've used elsewhere it to cover the c-style name as a deprecated symbol, with a message indicating how it has been renamed. Would that be reasonable?

Maybe for the event names I can see it but what other name do you have in mind for `pollAdd`?

jrflat:
My vote would be to keep `pollAdd`, change `pollEvents:` to just `events:`, rename the events to Guillaume's suggestion (`.readable`, `.writable`, `.error`, `.hangUp`, `.invalidDescriptor`, `.priorityData`, `.peerClosed`), and add the matching deprecated/renamed symbols with the original C names.

Sources/System/IORing/PollEvents.swift
```diff
@@ -0,0 +1,118 @@
...
+    ///
+    /// ```swift
+    /// // Monitor a socket for incoming data
+    /// let request = IORing.Request.pollAdd(
+    ///     socketFD,
+    ///     pollEvents: .pollIn,
+    ///     isMultiShot: true
+    /// )
+    /// ```
+    public struct PollEvents: OptionSet, Hashable, Codable, CaseIterable {
```
jrflat: Does it make sense for an `OptionSet` to be `CaseIterable` like this? It seems odd that a valid combination like `[.pollIn, .pollOut]` wouldn't be included in `.allCases`. Could we maybe just drop the `CaseIterable` and the `enum Event` and expose a `public static let individualEvents = [...]` if needed?

Sources/System/IORing/PollEvents.swift
```diff
@@ -0,0 +1,118 @@
...
+        public var rawValue: UInt32
```
jrflat: I think most of our `OptionSet`s use a `let` since `var` isn't needed for its mutating methods.

Sources/System/IORing/RawIORequest.swift
```diff
@@ -80,6 +84,27 @@ extension RawIORequest {
...
+    @_alwaysEmitIntoClient var pollEvents: IORing.Request.PollEvents {
+        get { .init(rawValue: _applyPollMask(rawValue.poll32_events)) }
+        set { rawValue.poll32_events = _applyPollMask(newValue.rawValue) }
+    }
+
+    /// Converts a poll event mask between its in-memory and `sqe` encodings.
+    ///
+    /// This is a halfword rotate (`swahw32` from `<linux/swab.h>`), which keeps
+    /// the byte order within each half.
+    @_alwaysEmitIntoClient func _applyPollMask(_ mask: UInt32) -> UInt32 {
```
jrflat: Could this be `static`?

Tests/SystemTests/IORingTests.swift
```diff
@@ -300,16 +302,314 @@ final class IORingTests: XCTestCase {
...
+        }
+    }
+
+    func testPollHangup() throws {
+        try XCTSkipIf(!uringEnabled, failureMessage)
+        var ring = try IORing(queueDepth: 8)
+        let (readFD, writeFD) = try FileDescriptor.pipe()
+        defer {
+            try? readFD.close()
+            try? writeFD.close()
```
jrflat: Could this get into very rare trouble if `writeFD` gets reused after closing below?

Sources/System/IORing/PollEvents.swift
FranzBusch: I removed the `CaseIterable` and renamed the property to `var allEvents`. I kept it a `var` because computed var's produce better performance compared to static let.

Sources/System/IORing/PollEvents.swift
FranzBusch: I changed it to `let`. For us in our server libraries we normally make all properties on a `struct` as `var`s by default since the value-type nature of the struct is already enforcing the mutability so if someone wants to store it in a `var` then they can change the raw value but since it is not super critical I changed it to what the project here prefers.

FranzBusch:
@glessard @jrflat @Catfish-Man I updated PR trying to address all the open comments. I left the naming as `pollAdd` since I still believe that is strictly better than using either `poll` or just `add`.

PR #278, opened 2026-01-07
Require Swift 6.1

glessard:
Whenever we determine that our next minor version will be cut after the release of Swift 6.3, then we should merge this.


jrflat (approved):
`release/1.7.0` is open if we want to target this there

FranzBusch:
> I read this to mean that once 6.3 is out we can drop 6.0.x; this is what we have done for the other packages.

Yes that's correct. Sorry I mixed up the versions about. If 6.3 is out then packages should support 6.1x., 6.2.x and 6.3.x

PR #279, opened 2026-01-07
Fix silent buffer size truncation in Windows pread/pwrite

mairinkdev:
### Problem

The Windows implementations of `pread` and `pwrite` in `WindowsSyscallAdapters.swift` convert `Int` buffer sizes to `DWORD` (32-bit unsigned integer) without validation. When `nbyte > DWORD.max` (4,294,967,295 bytes), the conversion silently truncates, causing:

- **Silent truncation**: Only a portion of the buffer is read/written
- **No error indication**: Functions return truncated count without signaling failure  
- **Data loss**: Callers may assume full buffer was processed
- **Undefined behavior**: Truncated value may be invalid for Windows API

### Solution

Add validation before the `DWORD` conversion to check if `nbyte > DWORD.max`. If exceeded, set `errno` to `EINVAL` and return `-1`, providing clear error indication instead of silent truncation.

### Changes

- **Sources/System/Internals/WindowsSyscallAdapters.swift**: Added buffer size validation in `pread` and `pwrite` functions
- **Tests/SystemTests/FileOperationsTestWindows.swift**: Added `testBufferSizeLimit()` to verify buffers exceeding `DWORD.max` return `EINVAL`

### Testing

- Added test case that verifies `read(fromAbsoluteOffset:into:)` and `write(toAbsoluteOffset:_:)` properly reject buffer sizes > 4GB with `EINVAL`
- Test uses a small buffer but passes oversized count to verify validation without allocating 4GB of memory

### Impact

- **No API changes**: Internal validation only, no public API modifications
- **Windows-only**: Does not affect other platforms
- **Backward compatible**: Existing code with valid buffer sizes is unaffected
- **Correctness improvement**: Prevents silent data loss in edge cases

### Why This Matters

This is a correctness bug that could affect real-world applications performing large file I/O operations on Windows. The fix ensures the library fails fast with a clear error rather than silently corrupting data.

### DIFF
--- a/Sources/System/Errno.swift
+++ b/Sources/System/Errno.swift
@@ -160,7 +160,7 @@ public struct Errno: RawRepresentable, Error, Hashable, Codable {
   /// No child processes.
   ///
   /// A `wait(2)` or `waitpid(2)` function was executed
-  /// by a process that dosn't have any existing child processes
+  /// by a process that doesn't have any existing child processes
   /// or whose child processes are all already being waited for.
   ///
   /// The corresponding C error is `ECHILD`.
--- a/Sources/System/Internals/WindowsSyscallAdapters.swift
+++ b/Sources/System/Internals/WindowsSyscallAdapters.swift
@@ -149,6 +149,12 @@ internal func pread(
   let handle: intptr_t = _get_osfhandle(fd)
   if handle == /* INVALID_HANDLE_VALUE */ -1 { ucrt._set_errno(EBADF); return -1 }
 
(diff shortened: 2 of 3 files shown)


glessard:
It would be preferable to have wrappers for the native syscalls instead of these specific POSIX-compatibility adaptors. This being said, this fix seems correct, and the error code is appropriate per common man pages. In particular EINVAL for "the resulting file offset would be negative, or beyond the end of a seekable device" (Linux) or "The specified file offset is invalid" (Darwin.)

glessard:
Thanks @mairinkdev! This is included in the 1.6.4 tag.

Issue #288, opened 2026-01-21
Consuming functions of resilient types cannot be inlinable


glessard:
Some of the consuming methods of `MachPort` were marked `@inlinable`, but it turns out this is not correct for a resilient type. `MachPort` is resilient on Darwin. (Discussed in rdar://160815058.)

PR #289, opened 2026-01-21
MachPort inlinability fixes

glessard:
When compiled with library resilience, `consuming` functions which `discard self` cannot also be `@inlinable`. While this does not cause issues in the package (it is not compiled with resilience on,) this is a back-port of an internal bug fix. A diagnostic for this error is introduced in Swift 6.3.

Addresses https://github.com/apple/swift-system/issues/288


glessard:
@swift-ci please test

Issue #188, opened 2024-07-15
Surprising behavior on Windows with drive roots in the middle of a path


jakepetroules:
Consider the following (tested with v1.3.1 tag):

```swift
let fp = FilePath("Sources/E:\\source")
print(fp.isAbsolute) // false
print(fp.isRelative) // true
print(fp.root) // nil
print(Array(fp.components)) // ["Sources", "E:", "source"]

print(FilePath("Sources").appending(fp.components)) // SystemPackage/FilePathParsing.swift:351: Assertion failed
print(FilePath("Sources").pushing(fp)) // Sources\Sources\E:\source

print(FilePath("Sources").appending(FilePath.Component("C:"))) // SystemPackage/FilePathString.swift:299: Fatal error: FilePath.Component must be created from exactly one non-root component
print(FilePath("Sources").appending("C:")) // Sources
```

It's not clear to me that some of these assertions are expected, or even how a seemingly bogus path like `Sources/E:\\source` _should_ be treated by the APIs. Should the FilePath constructor have asserted right away on this path if containing prohibited characters per https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions? Should there be an alternative validating constructor that returns nil or throws in this case?

lorentey:
Trapping in `FilePath.appending` does not seem right to me. Trapping in an _assertion_ (as opposed to precondition) failure is definitely incorrect -- this operation either needs to succeed, or it must produce a guaranteed runtime error.

If `Source/E:\\sources` is not a valid path on Windows, then I think `FilePath("Sources/E:\\source")` should not successfully create a file path. `FilePath.init(_: String)` is a non-failable initializer, so the only way to report an error is to trap. If that isn't acceptable, then we have to make a choice between deprecating it, or coming up (and clearly documenting!) a mechanism to turn it into a valid path.

For what it's worth, `Foo\0Bar` is clearly an invalid path on UNIX-compatible systems, but `FilePath("Foo\0Bar")` appears to accept it, silently truncating the input string before the NUL character. If we take this (terrible) precedent as the intended behavior, then one potential solution on Windows would be to filter out inappropriate characters or to replace them with some bogus substitute.

The state of `FilePath` makes me deeply unhappy.

weissi:
[...]

PR #302, opened 2026-05-25
Preserve leading `..` for drive-relative paths during lexical normalization

adityasingh2400:
On Windows, a traditional drive-relative root like `C:` anchors a path to the current working directory *on* the named drive, not to a fixed location. That makes a leading `..` meaningful: `C:..` is the parent of the drive's current directory, which is generally not the root of the drive. `lexicallyNormalize()` was treating any root as a fixed anchor and collapsing leading `..` against it, so it silently changed the meaning of drive-relative paths:

```swift
FilePath(#"C:..\foo\bar"#).lexicallyNormalized()    // produced C:foo\bar
FilePath(#"C:..\..\foo"#).lexicallyNormalized()     // produced C:foo
FilePath(#"C:foo\..\..\bar"#).lexicallyNormalized() // produced C:bar
```

In each case a `..` that escapes the relative portion was dropped, which loses information.

The root cause is in `_normalizeSpecialDirectories`: the branch that skips a leading `..` fired whenever the path had a root (`hasRoot && writeIdx == relStart`). That is correct for fixed anchors (`/` on Unix, and `\`, `C:\`, UNC and device roots on Windows, where `..` at the front refers to the parent of the root and collapses to the root), but it is wrong for the drive-relative `C:` form, where the `..` has to be kept just like a leading `..` on a rootless relative path.

The fix distinguishes the two cases. A new internal `Root._isTraditionalDriveRelative` identifies the `X:` form (a two-character root ending in `:`, which the existing `isAbsolute` logic already treats as the only colon-terminated relative root). Normalization now preserves a leading `..` exactly when the root, if any, is drive-relative. `isLexicallyNormal` is updated to match, which resolves the FIXME that was sitting on it: `C:..\foo\bar` is now reported as lexically normal, while `\..\foo\bar` is not.

After the change:

```swift
FilePath(#"C:..\foo\bar"#).lexicallyNormalized()    // C:..\foo\bar
FilePath(#"C:..\..\foo"#).lexicallyNormalized()     // C:..\..\foo
FilePath(#"C:foo\..\..\bar"#).lexicallyNormalized() // C:..\bar
FilePath(#"\..\foo\bar"#).lexicallyNormalized()    // \foo\bar  (unchanged)
```

These match `ntpath.normpath` for the same inputs (CPython recently corrected its own drive-relative normalization for the same reason in python/cpython#126780). The rooted `\` and absolute cases, and all Unix behavior such as `/.. => /`, are unaffected.

Verification: reproduced the wrong output against the documented/reference behavior using the package's `withWindowsPaths` test harness on macOS, applied the fix, and confirmed the outputs match. Added regression cases in `FilePathSyntaxTest` covering leading and interior `..` for both drive-relative (`C:`) and rooted (`\`) Windows paths; they fail before the change and pass after. `swift test` passes in both debug and release (67 XCTest tests and 7 swift-testing tests, 0 failures), and one pre-existing case that encoded the old behavior (`C:foo\bar\..\..\.. => C:`) was corrected to `C:..`.

This is related to the drive-root discussion in #188 (the trapping/validation questions there are separate and not addressed here).

I have not yet completed the Swift project CLA; happy to sign it as part of the review process per CONTRIBUTING.

### DIFF
--- a/Sources/System/FilePath/FilePathParsing.swift
+++ b/Sources/System/FilePath/FilePathParsing.swift
@@ -132,7 +132,11 @@ extension FilePath {
     defer { assert(isLexicallyNormal) }
 
     let relStart = _relativeStart
-    let hasRoot = relStart != _storage.startIndex
+
+    // Whether a leading `..` collapses against a fixed root. A traditional
+    // drive-relative root such as `C:` does not anchor leading `..`, so those
+    // are preserved exactly as for a rootless relative path.
+    let rootAnchorsLeadingParents = _rootAnchorsLeadingParents
 
     // TODO: all this logic might be nicer if _parseComponent considered
     // the null character index to be the next start...
@@ -153,8 +157,8 @@ extension FilePath {
       // otherwise parse-back a component to remove the parent (but stop at
       // root).
       if _isParentDirectory(component) {
-        // Skip over it if we're at the root
-        if hasRoot && writeIdx == relStart {
+        // Skip over it if it collapses against a fixed root
+        if rootAnchorsLeadingParents && writeIdx == relStart {
           readIdx = nextStart
           continue
         }
@@ -168,7 +172,7 @@ extension FilePath {
             readIdx = nextStart
             continue
           }
-          assert(self.root == nil && self.components.first!.kind == .parentDirectory)
+          assert(!rootAnchorsLeadingParents && self.components.first!.kind == .parentDirectory)
         }
       }
 
@@ -204,6 +208,16 @@ extension FilePath {
   }
   internal var _hasRoot: Bool {
     _relativeStart != _storage.startIndex
+  }
+
+  // Whether a leading `..` component refers to the parent of a fixed anchor
+  // (and therefore collapses during lexical normalization), as opposed to
+  // being preserved. This is true when the path has a root, unless that root
+  // is a traditional drive-relative root such as `C:`, which anchors to the
+  // drive's current directory rather than to a fixed location.
+  internal var _rootAnchorsLeadingParents: Bool {
+    guard let root = self.root else { return false }
+    return !root._isTraditionalDriveRelative
   }
 }
 
@@ -320,6 +334,21 @@ extension FilePath.Root {
       (slice.count == 1 && slice.first == .backslash) ||
         (slice.count == 2 && slice.last == .colon))
     return false
+  }
+
+  // Whether this is a traditional drive-relative root, i.e. `X:` with no
+  // trailing separator.
+  //
+  // Such a root anchors the path to the current working directory *on* the
+  // named drive rather than to a fixed directory, so a leading `..` is
+  // meaningful and must not be collapsed during lexical normalization
+  // (`C:..` is the parent of the drive's current directory). This is unlike
+  // `\` and `C:\`, where a leading `..` refers to the parent of a fixed root
+  // and collapses to the root itself.
+  internal var _isTraditionalDriveRelative: Bool {
+    guard _windowsPaths else { return false }
+    let slice = self._slice
+    return slice.count == 2 && slice.last == .colon
   }
 }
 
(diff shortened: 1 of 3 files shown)


milseman (approved):
This looks good to me. 

Note that `FilePath` is migrating to the stdlib, but System will still be providing lexical normalization so we'll want to make sure we keep this working. Thanks for the test cases!

milseman:
@swift-ci please test

adityasingh2400:
Rebased onto current main to get a fresh CI run.

The reds on the previous run were all runner-side rather than code: the macOS jobs failed at `xcrun --version` with `missing DEVELOPER_DIR path: /Applications/Xcode_16.3.app` before any build step ran, and the Windows job failed to reach the Docker daemon. Nothing in the build or test output was reached.

Locally on the rebased branch, `swift build` succeeds and `swift test` runs 70 tests with 0 failures and 2 skipped.

glessard:
`FilePath` is being moved to the standard library ([SE-0529](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0529-filepath-in-stdlib.md)). Please consider contributing the change there once the move has been completed.

adityasingh2400:
Thanks, I had not been tracking SE-0529.

I am happy to take this to the standard library once the move lands. Two questions so I do the right thing with this one:

Do you want it closed now, or left open as a pointer until the move is complete? I do not mind either way, and I would rather not leave a stale PR sitting in your queue if closing is cleaner for you.

Is the drive-relative `..` behaviour something the move is expected to carry over as is, or is normalization being revisited as part of it? If the semantics are in flux I would rather rebase the fix onto whatever lands than port the current one.

For reference, the change is small and self contained: `..` is preserved when it precedes a drive-relative root, since dropping it changes which directory the path resolves against. Tests are in `FilePathSyntaxTest.swift`. It is currently green and mergeable here if you would prefer to take it now and let the move carry it along.

glessard:
Many details of the Windows implementation are adjusted in the move to the standard library; the easiest way to evaluate the need once the move is complete would be to have a self-contained example that shows the behaviour being fixed here, either as a new issue or an update to an existing one. Let's not close the pull request yet.

adityasingh2400:
Here is the self-contained example.

```swift
import SystemPackage

for p in [#"C:.."#, #"C:..\foo\bar"#, #"C:..\..\foo"#,
          #"C:foo\..\..\bar"#, #"C:\.."#, #"\.."#] {
  print("\(p)  ->  \(FilePath(p).lexicallyNormalized())")
}
```

Observed against `main` at `1b452c2`, and against that same tree with only the two source files from this PR applied:

| input | main | with this PR |
|---|---|---|
| `C:..` | `C:` | `C:..` |
| `C:..\foo\bar` | `C:foo\bar` | `C:..\foo\bar` |
| `C:..\..\foo` | `C:foo` | `C:..\..\foo` |
| `C:foo\..\..\bar` | `C:bar` | `C:..\bar` |
| `C:\..` | `C:\` | `C:\` |
| `\..` | `\` | `\` |

The first four rows are the behaviour in question. `C:` is drive-relative, so it anchors to the current directory on drive C rather than to a fixed location, which makes a leading `..` meaningful. `cd C:..` in cmd.exe moves to the parent of that directory. Dropping it makes the normalized path name a different directory than the original.

The last two rows are the contrast and are unchanged here. `C:\` and `\` are fixed roots, so a leading `..` collapses correctly.

I produced those numbers on macOS rather than Windows, using the existing `withWindowsPaths(enabled: true)` hook, so it is the same parsing path the `.windows(...)` cases in `FilePathSyntaxTest` already exercise.

One further data point on whether the current behaviour is deliberate. With only this PR's two source files applied and the test file left at `main`, the suite fails at `FilePathSyntaxTest.swift:648` on `lexically normalized`, because `main` carries a case asserting that `C:..` normalizes to `C:`. With this PR's test file as well it is 70 tests, 2 skipped, 0 failures. So the existing behaviour is pinned by a test today, which seems worth knowing when the same question reaches the standard library.

Happy to open this as an issue against whichever repo is the right home once the move lands.

PR #307, opened 2026-05-29
Use fatalError instead of precondition for missing permissions on .create

### LINKED ISSUE #78: FileDescriptor.open(_:_:options:permissions:retryOnInterrupt:) should throw an error instead of trapping on unspecified permissions
`FileDescriptor.open(_:_:options:permissions:retryOnInterrupt:)` [traps](https://github.com/apple/swift-system/blob/1d5e35a460ebef822d61181b42a7354abaeb80c3/Sources/System/FileOperations.swift#L80) if `options.contains(.create)` and `permissions` is `nil`. since its callers can `throw`, why not throw an error instead?

alternatively, if this isn’t appropriate for `throws`, it should be a `fatalError(_:)` instead of a `precondition`, so users don’t have to hunt through source code to figure out why `FileDescriptor` is crashing…

adityasingh2400:
## Problem

`FileDescriptor.open(_:_:options:permissions:retryOnInterrupt:)` traps when
`options` contains `.create` but `permissions` is `nil`. As reported in #78,
this trap was raised with `precondition`, whose failure message can be elided
in optimized builds. When that happens the program aborts without any
indication of *why*, forcing users to dig through the source to discover that
`.create` requires a non-`nil` `permissions` value.

The reporter concluded this is a programmer error (so `throws` is not
appropriate), but that the crash should still carry a clear message in the
crash dump.

## Change

- Replace the `precondition(!options.contains(.create), ...)` check in the
  internal `_open` implementation with an explicit
  `if options.contains(.create) { fatalError(...) }`. Unlike `precondition`,
  `fatalError` always fires and always prints its message regardless of
  optimization level, so the diagnostic is guaranteed to appear in the crash
  dump.
- Use a descriptive message:
  `"FileDescriptor.open: 'permissions' must not be nil when 'options' contains '.create'"`.
- Document the trapping behavior in the `permissions:` parameter docs of the
  three public `open` overloads, so the requirement is discoverable from the
  API reference.

The change is confined to the existing trapping code path; no public API,
signatures, or non-trapping behavior change.

## Verification

- `swift build` succeeds (forced recompile of `FileOperations.swift`).
- `swift test --filter FileOperationsTest` passes: 8/8 tests.
- Full suite passes except `TemporaryPathTest.testNotInSlashTmp`, which is an
  environmental failure unrelated to this change (it asserts the temp dir is
  not under `/tmp`, and fails identically on an unmodified checkout when the
  test process runs from `/tmp`).

Fixes #78

### DIFF
--- a/Sources/System/FileOperations.swift
+++ b/Sources/System/FileOperations.swift
@@ -16,6 +16,8 @@ extension FileDescriptor {
   ///   - mode: The read and write access to use.
   ///   - options: The behavior for opening the file.
   ///   - permissions: The file permissions to use for created files.
+  ///     This value must not be `nil` when `options` contains `.create`;
+  ///     passing `nil` in that case is a programmer error and traps at runtime.
   ///   - retryOnInterrupt: Whether to retry the open operation
   ///     if it throws ``Errno/interrupted``.
   ///     The default is `true`.
@@ -55,6 +57,8 @@ extension FileDescriptor {
   ///   - mode: The read and write access to use.
   ///   - options: The behavior for opening the file.
   ///   - permissions: The file permissions to use for created files.
+  ///     This value must not be `nil` when `options` contains `.create`;
+  ///     passing `nil` in that case is a programmer error and traps at runtime.
   ///   - retryOnInterrupt: Whether to retry the open operation
   ///     if it throws ``Errno/interrupted``.
   ///     The default is `true`.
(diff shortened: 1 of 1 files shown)


milseman:
This looks ok to me in principle. We should also (separately) change the compiler/stdlib to preserve precondition messages. It's bizarre to drop them, but that's separate from this targeted fix.

milseman (approved):

milseman:
@swift-ci please test

glessard (approved):

PR #309, opened 2026-06-04
Fix a crash when work is interleaved between prepare and submit in IORing

Catfish-Man:


Sources/System/IORing/IORing.swift
```diff
@@ -41,6 +41,32 @@ extension UnsafeMutableRawBufferPointer {
     }
 }
 
+/// Owns the FilePaths whose interior pointers have been written into pending
+/// SQEs. The kernel copies SQE-referenced data out during io_uring_enter (due to
+/// IORING_FEAT_SUBMIT_STABLE), so the paths only need to live across the prepare->submit gap
+@usableFromInline
+internal final class PendingPathBuffers {
```
glessard: Can we include an explanation for this being a class?

glessard (approved):
We should target either the 1.6.0 or the 1.7.0 branch, depending on where we think this will land.

Sources/System/IORing/IORing.swift
```diff
@@ -41,6 +41,32 @@ extension UnsafeMutableRawBufferPointer {
...
+    init(reservedCapacity: Int) {
+        paths.reserveCapacity(reservedCapacity)
+    }
+
+    @usableFromInline
+    func pin(_ path: FilePath) -> UnsafePointer<CInterop.PlatformChar> {
+        paths.append(path)
+        // Safe to escape: `paths` keeps the FilePath (and its underlying
+        // [SystemChar] buffer) alive until clear() or class deinit.
+        return paths.last!.withPlatformString { $0 }
```
jrflat: Not blocking since this is functionally correct, but we could consider copying the bytes + null-terminator into our own `ContiguousArray<CInterop.PlatformChar>` and storing that for robustness against any change in the underlying `SystemString` representation. (There are some TODOs to refactor it to an interchangeable storage, and something like an inline storage could cause UAF.)

jrflat (approved):

Sources/System/IORing/IORing.swift
Catfish-Man: Yeah I thought about that, but I'm *really* trying to avoid mallocs here. We have tests for this so if we change that, we should be able to change this as well in response.

PR #312, opened 2026-06-09
Add `pipe(options:)` for Windows

glessard:
This adds:
- `FileDescriptor.PipeOptions` for Windows, enabling the option `.closeOnExec`.
- `FileDescriptor.pipe(options:)` for Windows

...along with some extra tests.

The Windows-native implementation of pipes, using `CreatePipe()`, creates file handles that aren't inherited. Accordingly, the Windows `pipe()` in swift-system sets the no-inherit option. Adding a version of pipe() that accepts options gives us the opportunity to allow Windows users to not set the option when that is appropriate to their use case.

### DIFF
--- a/Proposals/0007-FileDescriptor-dup3-pipe2.md
+++ b/Proposals/0007-FileDescriptor-dup3-pipe2.md
@@ -167,7 +167,7 @@ struct FileDescriptor {
 }
 ```
 
-These API additions are unavailable Darwin, as the underlying `dup3` and `pipe2` APIs do not exist.
+These API additions are unavailable on Darwin, as the underlying `dup3` and `pipe2` APIs do not exist.
 
 `pipe(options:)` is added for Windows with the option `.closeOnExec`, but `duplicate(as:options:)` is not added because the underlying API does not match `dup3`.
 
--- a/Proposals/0007-FileDescriptor-dup3-pipe2.md
+++ b/Proposals/0007-FileDescriptor-dup3-pipe2.md
@@ -9,6 +9,8 @@
 
 * **v1** Initial version
 
+- **v2** Add Windows version of `pipe(options:)`
+
 ## Introduction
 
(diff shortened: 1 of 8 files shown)


grynspan (approved):

Proposals/0007-FileDescriptor-dup3-pipe2.md
```diff
@@ -165,7 +167,9 @@ struct FileDescriptor {
 }
 ```
 
-These API additions are unavailable on Windows and Darwin, as the underlying `dup3` and `pipe2` APIs do not exist.
+These API additions are unavailable Darwin, as the underlying `dup3` and `pipe2` APIs do not exist.
```
glessard:
```suggestion
These API additions are unavailable on Darwin, as the underlying `dup3` and `pipe2` APIs do not exist.
```

Sources/System/FileOperations.swift
jrflat, line 544: Can we add a note in the doc-comment explaining how this API does not include `_O_NOINHERIT` by default, so `pipe(options: [])` is not equivalent to `pipe()` on Windows?

jrflat (approved):

Proposals/0007-FileDescriptor-dup3-pipe2.md
grynspan: `dup3()` and `pipe2()` are available on anyAppleOS 27!

Sources/CSystem/include/CSystemShared.h
```diff
@@ -7,5 +7,8 @@
  See https://swift.org/LICENSE.txt for license information
 */
 
+#if !defined(_WIN32)
```
jakepetroules, line 10: didn't you mean to surround dup3 with `#if !defined(_WIN32)` rather than pipe2?

glessard: There is a Windows-specific implementation of `csystem_posix_pipe2` in WindowsSyscallAdapters.swift. The `dup3` stub is still relevant for Windows.

jakepetroules:
Actually, `open` supports `_O_NOINHERIT`, too. We should add it there as well for consistency.

https://github.com/apple/swift-system/pull/331

Issue #316, opened 2026-06-12
1.7.0 tag fails to compile on FreeBSD 14.x


glessard:
The `O_CLOFORK` flag was introduced in FreeBSD 15.0.

Unfortunately the FreeBSD toolchain doesn't have support for availability annotations. We could hard-code the value from the FreeBSD header, and document that using the flag on earlier versions will cause an error.

glessard:
@kkebo do you think hard-coding the flag value would be an acceptable compromise for FreeBSD 14.x?

kkebo:
@glessard I commented https://github.com/apple/swift-system/pull/317#issuecomment-4691782878.

glessard:
Fixed by #319.

PR #317, opened 2026-06-12
Circumvent FreeBSD’s closeOnFork definition issue

glessard:
Potential solution for https://github.com/apple/swift-system/issues/316

FreeBSD defined `O_CLOFORK` in version 15.0, but we test on FreeBSD 14.3.
This hard-codes the value from the header when the compilation target is FreeBSD.


glessard:
cc @kkebo

kkebo:
I have another idea. #319

glessard:
Excellent! I'd thought of trying that overnight =)

PR #318, opened 2026-06-12
fix: build with MemberImportVisibility

vsarunas:
Fixes build with:
```
swift build -Xswiftc -enable-upcoming-feature -Xswiftc MemberImportVisibility
[...]
FilePathTempPosix.swift:95:23: error: property 'd_name' is not available due to missing import of defining module 'Darwin' [#MemberImportVisibility]
 11 |
 12 | /// Get the path to the system temporary directory.
 13 | internal func _getTemporaryDirectory() throws -> FilePath {
    | `- note: add import of module 'Darwin'
 14 |   guard let tmp = system_getenv("TMPDIR") else {
 15 |     return "/tmp"
    :
 93 |   while let dirent = system_readdir(dir) {
 94 |     // Skip . and ..
 95 |     if dirent.pointee.d_name.0 == 46
    |                       `- error: property 'd_name' is not available due to missing import of defining module 'Darwin' [#MemberImportVisibility]
 96 |          && (dirent.pointee.d_name.1 == 0
 97 |                || (dirent.pointee.d_name.1 == 46
```

Builds fail on projects which set  `MemberImportVisibility` and import swift-system:
```
/home/actions/_work/ordo/ordo/.build/checkouts/swift-nio/Sources/_NIOFileSystem/Internal/System Calls/FileDescriptor+Syscalls.swift:75:31: error: initializer 'init()' is not available due to missing import of defining module 'CSystem' [#MemberImportVisibility]
 21 | @preconcurrency import Glibc
 22 | import CNIOLinux
 23 | #elseif canImport(Musl)
    | `- note: add import of module 'CSystem'
 24 | @preconcurrency import Musl
 25 | import CNIOLinux
    :
 73 |     @_spi(Testing)
 74 |     public func status() -> Result<CInterop.Stat, Errno> {
 75 |         var status = CInterop.Stat()
    |                               `- error: initializer 'init()' is not available due to missing import of defining module 'CSystem' [#MemberImportVisibility]
 76 |         return nothingOrErrno(retryOnInterrupt: false) {
 77 |             system_fstat(self.rawValue, &status)
```

Could you tag a new patch release @glessard  / @jrflat ?

### DIFF
--- a/Sources/System/FilePath/FilePathTempPosix.swift
+++ b/Sources/System/FilePath/FilePathTempPosix.swift
@@ -9,6 +9,23 @@
 
 #if !os(Windows)
 
+#if SYSTEM_PACKAGE_DARWIN
+import Darwin
+#elseif canImport(Glibc)
+import CSystem
+import Glibc
+#elseif canImport(Musl)
+import CSystem
+import Musl
+#elseif canImport(WASILibc)
+import WASILibc
+#elseif canImport(Android)
+import CSystem
+import Android
+#else
+#error("Unsupported Platform")
+#endif
+
 /// Get the path to the system temporary directory.
 internal func _getTemporaryDirectory() throws -> FilePath {
   guard let tmp = system_getenv("TMPDIR") else {


Sources/System/FilePath/FilePathTempPosix.swift
```diff
@@ -9,6 +9,23 @@
...
+#if SYSTEM_PACKAGE_DARWIN
+import Darwin
+#elseif canImport(Glibc)
+import CSystem
+import Glibc
+#elseif canImport(Musl)
+import CSystem
+import Musl
+#elseif canImport(WASILibc)
+import WASILibc
```
jrflat, line 21: I think we'll need the import for WASI, too

jrflat (commented):
Thanks for addressing this! The PR does address the first issue when building `swift-system` with `MemberImportVisibility`. However, I think the second issue (the `swift-nio` build) will require more changes on System side. Looks like the public `CInterop.Stat = stat` typealias is attributed to the internal `CSystem` module, requiring downstream consumers to `import CSystem`. We're looking into a fix!

vsarunas:
Yes looks like a change in NIO is also needed!

glessard:
@swift-ci please test

glessard (approved):

glessard:
This has been cherry-picked into https://github.com/apple/swift-system/releases/tag/1.7.2

PR #319, opened 2026-06-12
Fix missing `O_CLOFORK` on older FreeBSD

### LINKED ISSUE #316: 1.7.0 tag fails to compile on FreeBSD 14.x

kkebo:
fixes #316 

closes #317 

In FreeBSD, `O_CLOFORK` has been added in https://github.com/freebsd/freebsd-src/pull/1698. This change hasn't ported to FreeBSD 14.x, so FreeBSD 14.x CI is currently failing.

This PR added a new variable `FREEBSD_O_CLOFORK`, which wraps `O_CLOFORK` on newer FreeBSD and `0` on older FreeBSD, and resolved #316 by using it.

### DIFF
--- a/Sources/CSystem/CMakeLists.txt
+++ b/Sources/CSystem/CMakeLists.txt
@@ -16,6 +16,7 @@ install(FILES
   include/CSystemLinux.h
   include/CSystemShared.h
   include/CSystemWindows.h
+  include/CSystemFreeBSD.h
   include/module.modulemap
   DESTINATION include/CSystem)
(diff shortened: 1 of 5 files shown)


glessard (approved):
Excellent solution

PR #320, opened 2026-06-12
[1.7.0] Fix Stat MemberImportVisibility regression and add test

jrflat:
With `MemberImportVisibility` enabled, consumers of `SystemPackage` that read C `stat` fields fail to build due to the public `CInterop.Stat = stat` typealias (such as reported in https://github.com/apple/swift-nio/issues/3622):

```
- error: property 'st_mode' is not available due to missing import of defining module 'CSystem' [#MemberImportVisibility]
``` 

`CSystem` re-exported `<sys/stat.h>` and `<fcntl.h>`, so `struct stat`'s members were attributed to the private `CSystem` module rather than the platform overlay. `CSystem` uses neither header, so we can remove both includes from `CSystemLinux.h`, and `struct stat` will now resolve through Glibc/Musl/Bionic, which consumers already import.

To prevent this sort of regression in the future, this PR also adds a compile-only `MemberImportVisibility` test target that touches all `CInterop.Stat` members, so `swift test` should catch any issues in CI. Future wrapper types introduced to System can be added there, too.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -141,6 +141,12 @@ let package = Package(
       exclude: testsToExclude,
       cSettings: cSettings,
       swiftSettings: swiftSettings),
+    // Builds a SystemPackage consumer with MemberImportVisibility enabled.
+    // Fails if a C member of a vended type is attributed to the CSystem module.
+    .testTarget(
+      name: "MemberImportVisibility",
+      dependencies: ["SystemPackage"],
+      swiftSettings: [.enableUpcomingFeature("MemberImportVisibility")]),
   ],
   swiftLanguageModes: [.v5]
 )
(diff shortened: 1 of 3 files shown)


jrflat:
Android build of the canary probe is failing with e.g:

```
- error: property 'st_atim' is not available due to missing import of defining module 'posix_filesystem' [#MemberImportVisibility]
```

due to its modularized libc headers. I think this means the real value of the canary build is only for Glibc/Musl where libc headers are un-modularized and `CSystem` can shadow `struct stat`, so I'm going to remove Android from the test here.

glessard (approved):

finagolfin:
> I think this means the real value of the canary build is only for Glibc/Musl where libc headers are un-modularized and CSystem can shadow struct stat, so I'm going to remove Android from the test here.

More likely because the `Bionic` submodule doesn't work and was botched from the beginning. Can you try it with `canImport(Android)` instead?

PR #325, opened 2026-06-13
Add missing System sources to CMake build

fallintoplace:
## Summary

This updates the CMake source list for SystemPackage so it matches the Swift sources in the SwiftPM target.

The CMake build now includes the previously omitted source groups:

- Windows support files
- temporary-path implementation files
- FileSystem APIs, including Stat, FileType, FileMode, and FileFlags
- Linux-only IORing files

It also enables the Swift Lifetimes feature for the CMake target, matching SwiftPM, so the Linux IORing sources are not compiled away when they are included.

## Why

The CMake target had drifted from the SwiftPM target. CMake consumers could get a successful build while the produced module was missing newer API areas.

## Validation

- cmake configure with Ninja and CMAKE_INSTALL_PREFIX=.build/install-system-sources-check
- cmake --build .build/cmake-system-sources-check
- cmake --install .build/cmake-system-sources-check
- Installed-module Swift typecheck that imports SystemPackage and uses FilePath.stat() / Stat.type
- Source-list audit: CMake now names 40/40 Swift sources from the SwiftPM target

### DIFF
--- a/Sources/System/CMakeLists.txt
+++ b/Sources/System/CMakeLists.txt
@@ -9,6 +9,7 @@ See https://swift.org/LICENSE.txt for license information
 
 add_library(SystemPackage
   Errno.swift
+  ErrnoWindows.swift
   FileDescriptor.swift
   FileHelpers.swift
   FileOperations.swift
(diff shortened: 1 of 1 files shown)


jrflat (commented):
LGTM, thanks! I verified locally on macOS/Linux, but I'll also trigger cross-repo CI that exercises the CMake path to confirm before merging

fallintoplace:
Do you still need anything from my side?

jrflat (approved):
Thanks! Finally got a successful Ubuntu 24.04 toolchain build in CI as the last bit of verification.

PR #328, opened 2026-06-26
Don't define __BSD_VISIBLE on OpenBSD.

3405691582:
This causes a compilation error because OpenBSD headers that refer to this aren't `#ifdef` but `#if`, so the define must have a value.

This doesn't actually seem to be necessary on OpenBSD, so let's just remove it on OpenBSD.

### DIFF
--- a/Sources/CSystem/shims.c
+++ b/Sources/CSystem/shims.c
@@ -7,10 +7,10 @@
  See https://swift.org/LICENSE.txt for license information
 */
 
-#if defined(__FreeBSD__) || defined(__OpenBSD__)
 #if defined(__FreeBSD__)
 #define __BSD_VISIBLE
 #endif
+#if defined(__FreeBSD__) || defined(__OpenBSD__)
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


Sources/CSystem/shims.c
```diff
@@ -8,7 +8,9 @@
 */
 
 #if defined(__FreeBSD__) || defined(__OpenBSD__)
+#if defined(__FreeBSD__)
```
glessard: I feel like structuring this slightly differently would be easier to read:
```suggestion
#if defined(__FreeBSD__)
```

Sources/CSystem/shims.c
```diff
@@ -8,7 +8,9 @@
...
 #define __BSD_VISIBLE
+#endif
```
glessard, line 12: Along with:
```suggestion
#endif
#if defined(__FreeBSD__) || defined(__OpenBSD__)
```

glessard (commented):
Looks like a reasonable change.

glessard:
@swift-ci please test

PR #329, opened 2026-06-26
Test ioring's blocking consume completion with timeout

glessard:
This adds a test to exercise the timeout behaviour of `IORing`, fixing other issues along the way.

This comes from trying to exercise the fix in https://github.com/apple/swift-system/pull/326.


Catfish-Man (approved):
LGTM, thank you!

PR #331, opened 2026-06-27
Enable .closeOnExec for FileDescriptor.open on Windows

jakepetroules:
Make OpenOptions.closeOnExec available on Windows, where it maps to _O_CLOEXEC (O_NOINHERIT). The Windows open adapter already decodes O_NOINHERIT into a non-inheritable handle, so this produces the expected close-on-exec semantics. Also add the deprecated O_NOINHERIT alias, matching the PipeOptions treatment from #312.

### DIFF
--- a/Sources/System/FileDescriptor.swift
+++ b/Sources/System/FileDescriptor.swift
@@ -301,7 +301,6 @@ extension FileDescriptor {
     public static var O_EVTONLY: OpenOptions { eventOnly }
 #endif
 
-#if !os(Windows)
     /// Indicates that executing a program closes the file.
     ///
     /// Normally, file descriptors remain open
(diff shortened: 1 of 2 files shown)


jakepetroules:
@swift-ci test

Sources/System/FileDescriptor.swift
```diff
@@ -319,6 +318,11 @@ extension FileDescriptor {
     @_alwaysEmitIntoClient
     @available(*, unavailable, renamed: "closeOnExec")
     public static var O_CLOEXEC: OpenOptions { closeOnExec }
```
grynspan, line 320: Should probably still be `!Windows` here.

jakepetroules: I'm just staying consistent with what PipeOptions is doing.

grynspan: I _think_ that's just a typo in `PipeOptions` that we didn't catch.

glessard: The point is to help someone who would be aware of the functionality but tries to use it with the wrong spelling.

glessard:
@swift-ci please test

jrflat (approved):

glessard:
@swift-ci please test

PR #336, opened 2026-07-06
Improve syscall wrappers for IORing

glessard:


Catfish-Man (approved):
Much nicer, thank you!

Sources/System/IORing/IORing.swift
```diff
@@ -731,15 +716,12 @@ public struct IORing: ~Copyable {
 
     /// Removes registered files from the ring
     public func unregisterFiles() throws {
```
jrflat, line 718: Do we want to try to make this `throws(Errno)` while we're at it?

Sources/System/IORing/IORing.swift
```diff
@@ -805,15 +782,12 @@ public struct IORing: ~Copyable {
     }
 
     public func unregisterBuffers() throws {
```
jrflat, line 784: Also could be tightened to `throws(Errno)`

jrflat (approved):
LGTM left some non-blocking comments on functions we could probably tighten to `throws(Errno)`

Sources/System/IORing/IORing.swift
glessard: Good idea; let's handle this in a different PR, since it's quite unrelated.

PR #339, opened 2026-07-09
[ci] validate cmake build on Linux

glessard:
Adds a CI job to validate that a change won't prevent CMake from building the package.

Adapted from https://github.com/swiftlang/swift-foundation/pull/1366

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -99,11 +99,12 @@ jobs:
           jq '.targets[] | select(.name == "SystemPackage") | .sources[]' /tmp/swiftpm.json | sort > /tmp/swiftpm-list.txt
           jq '.sources[].path[15:]' /tmp/cmake.json | sort > /tmp/cmake-list.txt
           comm -23 /tmp/swiftpm-list.txt /tmp/cmake-list.txt > /tmp/missing.txt
-          test ! -s /tmp/missing.txt
-          exit_code=$?
-          test $exit_code -ne 0 && echo "::notice::SystemPackage CMakeLists.txt matches SwiftPM"
-          test $exit_code -eq 0 || echo "::error::SystemPackage CMakeLists.txt does not match SwiftPM"
-          exit $exit_code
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -75,6 +75,39 @@ jobs:
...
+          cmake -G 'Ninja' -B build -S . -DCMAKE_C_COMPILER=clang -DCMAKE_Swift_COMPILER=swiftc -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=YES
+      - name: Validate CMake File Lists
+        shell: bash
+        run: |
+          swift package describe --type json > /tmp/swiftpm.json
+          cat build/.cmake/api/v1/reply/target-SystemPackage-*.json > /tmp/cmake.json
+          jq '.targets[] | select(.name == "SystemPackage") | .sources[]' /tmp/swiftpm.json | sort > /tmp/swiftpm-list.txt
+          jq '.sources[].path[15:]' /tmp/cmake.json | sort > /tmp/cmake-list.txt
+          comm -23 /tmp/swiftpm-list.txt /tmp/cmake-list.txt > /tmp/missing.txt
+          test ! -s /tmp/missing.txt
```
jrflat: `shell: bash` in GitHub actions corresponds to `bash --noprofile --norc -eo pipefail {0}`, so `-e` will fail the script immediately if there's missing files and I don't think error line below will be printed. Maybe we could use something like this?

```
if [ -s /tmp/missing.txt ]; then
  echo "::error::SystemPackage CMakeLists.txt does not match SwiftPM. Missing sources:"
  cat /tmp/missing.txt
  exit 1
fi
echo "::notice::SystemPackage CMakeLists.txt matches SwiftPM"
```

glessard: That explains why my "echo" statements weren't working. I took your version and both branches work as intended.

jrflat (approved):
The end of missing CMake sources! 🎉

PR #340, opened 2026-07-09
Fix a test reliability issue

glessard:
The timestamps around file modifications in `StatTests.times()` are frequently equal on Android, causing spurious failures. This changes the comparison from `>` to `>=`.


jrflat (approved):
Thanks!

PR #342, opened 2026-07-10
IORing throwing tweaks

glessard:
This tweaks two public IORing functions to specify error types.
One internal helper function is changed to have a generic typed error instead of having a `rethrows` annotation.


glessard:
@swift-ci please test

jrflat (approved):
Nice!

PR #343, opened 2026-07-10
Set -package-name in CMake build

owenv:
Currently the CMake build of main fails on Windows because it recently started using the package access level for the first time, but doesn't set a package name. Update CMakeLists.txt to set one.


owenv:
@swift-ci test

glessard:
@owenv The commit where the `package var` declarations were introduced is also on the release/1.7.0 branch. Could you target your PR to that branch instead?

glessard (approved):

glessard:
Thanks @owenv !

PR #348, opened 2026-07-15
Fix 64-bit file offsets on Windows

YayoRazo:
## Summary

On Windows, C `off_t` is `long`, i.e. 32-bit, so it cannot represent file offsets or sizes at or beyond 2 GiB. The internal offset typealias `_COffT` was unconditionally `off_t`, so `FileDescriptor.seek`, `read`/`write` at an absolute offset, and `resize` funneled a public `Int64` through a 32-bit conversion.

Because Swift's numeric conversions are range-checked:

- Any offset `>= 2^31` **trapped at runtime** (process crash), so seeking, positioned reads/writes, or resizing anywhere past 2 GiB was impossible.
- Even below the trap boundary, positioned I/O could never target a region past 4 GiB, because the `OVERLAPPED` high dword was computed from an already-32-bit value and was therefore always zero.

The Windows adapters already contained a 64-bit `lseek` overload backed by `_lseeki64` and an `OVERLAPPED`/`LARGE_INTEGER` split, which is strong evidence the 64-bit path was intended and only the plumbing type was wrong.

## Fix

- Make `_COffT` 64-bit on Windows (`Int64`); it remains `off_t` on all other platforms, so there is no behavioral change off Windows.
- Thread `_COffT` through `system_lseek`, `system_pread`, `system_pwrite`, and `system_ftruncate`.
- Remove the now-redundant 32-bit `_lseek` overload so overload resolution selects the existing `_lseeki64`-backed one.
- Use truncating conversions for the `OVERLAPPED` offset split, since `DWORD(UInt32(offset))` would itself trap once offsets exceed 4 GiB.

## Testing

- Added `testLargeFileOffsets` (Windows), which seeks past 2 GiB and 4 GiB and performs a positioned write/read round-trip beyond 4 GiB using a sparse file so it allocates no real storage.
- Verified the test **fails without this change** (it traps at the first seek past `Int32.max`) and passes with it.
- Full `swift test` suite green on Windows (`x86_64-unknown-windows-msvc`, Swift 6.3.3); no changes to other platforms.

### DIFF
--- a/Sources/System/Internals/Exports.swift
+++ b/Sources/System/Internals/Exports.swift
@@ -32,7 +32,16 @@ import Android
 #error("Unsupported Platform")
 #endif
 
+#if os(Windows)
+// On Windows, C `off_t` is `long`, i.e. 32-bit, so it cannot represent file
+// offsets or sizes at or beyond 2 GiB. Use a 64-bit offset type so that
+// seeking, positioned I/O, and resizing work on large files, matching the
+// 64-bit `off_t` on Linux and Darwin. The Windows syscall adapters route
+// this through the 64-bit `_lseeki64` / `OVERLAPPED` / `LARGE_INTEGER` APIs.
+internal typealias _COffT = Int64
+#else
 internal typealias _COffT = off_t
+#endif
 
 // MARK: syscalls and variables
 
--- a/Sources/System/Internals/WindowsSyscallAdapters.swift
+++ b/Sources/System/Internals/WindowsSyscallAdapters.swift
@@ -122,13 +122,6 @@ internal func write(
 }
 
 @inline(__always)
-internal func lseek(
-  _ fd: Int32, _ off: off_t, _ whence: Int32
-) -> off_t {
-  _lseek(fd, off, whence)
-}
-
-@inline(__always)
 internal func dup(_ fd: Int32) -> Int32 {
   _dup(fd)
 }
@@ -144,7 +137,7 @@ internal func dup2(_ fd: Int32, _ fd2: Int32) -> Int32 {
 
 @inline(__always)
 internal func pread(
-  _ fd: Int32, _ buf: UnsafeMutableRawPointer!, _ nbyte: Int, _ offset: off_t
+  _ fd: Int32, _ buf: UnsafeMutableRawPointer!, _ nbyte: Int, _ offset: Int64
 ) -> Int {
   let handle: intptr_t = _get_osfhandle(fd)
   if handle == /* INVALID_HANDLE_VALUE */ -1 { ucrt._set_errno(EBADF); return -1 }
@@ -159,8 +152,10 @@ internal func pread(
   let hFile: HANDLE = HANDLE(bitPattern: handle)!
 
   var ovlOverlapped: OVERLAPPED = OVERLAPPED()
-  ovlOverlapped.OffsetHigh = DWORD(UInt32(offset >> 32) & 0xffffffff)
-  ovlOverlapped.Offset = DWORD(UInt32(offset >> 0) & 0xffffffff)
+  // Split the 64-bit offset into high/low DWORDs. Use truncating conversions:
+  // `DWORD(UInt32(offset))` would trap for any offset >= 4 GiB.
+  ovlOverlapped.OffsetHigh = DWORD(truncatingIfNeeded: offset >> 32)
+  ovlOverlapped.Offset = DWORD(truncatingIfNeeded: offset)
 
   var nNumberOfBytesRead: DWORD = 0
   if !ReadFile(hFile, buf, DWORD(nbyte), &nNumberOfBytesRead, &ovlOverlapped) {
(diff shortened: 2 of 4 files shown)


Sources/System/Internals/Exports.swift
```diff
@@ -32,7 +32,16 @@ import Android
 #error("Unsupported Platform")
 #endif
 
+#if os(Windows)
+// On Windows, C `off_t` is `long`, i.e. 32-bit, so it cannot represent file
+// offsets or sizes at or beyond 2 GiB. Use a 64-bit offset type so that
+// seeking, positioned I/O, and resizing work on large files, matching the
+// 64-bit `off_t` on Linux and Darwin. The Windows syscall adapters route
+// this through the 64-bit `_lseeki64` / `OVERLAPPED` / `LARGE_INTEGER` APIs.
+internal typealias _COffT = Int64
```
compnerd, line 41: I think I would slightly prefer `_off64_t` as the spelling. This is still `Int64` under the hood, but more explicit and uses the libc spelling.

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -144,7 +137,7 @@ internal func dup2(_ fd: Int32, _ fd2: Int32) -> Int32 {
 
 @inline(__always)
 internal func pread(
-  _ fd: Int32, _ buf: UnsafeMutableRawPointer!, _ nbyte: Int, _ offset: off_t
+  _ fd: Int32, _ buf: UnsafeMutableRawPointer!, _ nbyte: Int, _ offset: Int64
```
compnerd, line 140: These can then use the same spelling

Sources/System/Internals/Exports.swift
```diff
@@ -32,7 +32,16 @@ import Android
 #error("Unsupported Platform")
 #endif
 
+#if os(Windows)
+// On Windows, C `off_t` is `long`, i.e. 32-bit, so it cannot represent file
+// offsets or sizes at or beyond 2 GiB. Use a 64-bit offset type so that
+// seeking, positioned I/O, and resizing work on large files, matching the
+// 64-bit `off_t` on Linux and Darwin. The Windows syscall adapters route
+// this through the 64-bit `_lseeki64` / `OVERLAPPED` / `LARGE_INTEGER` APIs.
```
glessard, line 40: Could you make this comment less wordy? It's probably enough to say that we use a 64-bit value because `off_t` is a 32-bit type on Windows.

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -159,8 +152,10 @@ internal func pread(
   let hFile: HANDLE = HANDLE(bitPattern: handle)!
 
   var ovlOverlapped: OVERLAPPED = OVERLAPPED()
-  ovlOverlapped.OffsetHigh = DWORD(UInt32(offset >> 32) & 0xffffffff)
-  ovlOverlapped.Offset = DWORD(UInt32(offset >> 0) & 0xffffffff)
+  // Split the 64-bit offset into high/low DWORDs. Use truncating conversions:
+  // `DWORD(UInt32(offset))` would trap for any offset >= 4 GiB.
```
glessard, line 156: This comment isn't necessary. The truncating API is wordy for a reason!

Sources/System/Internals/WindowsSyscallAdapters.swift
```diff
@@ -187,8 +182,10 @@ internal func pwrite(
...
```
glessard, line 186: Same as above

Tests/SystemTests/FileOperationsTestWindows.swift
```diff
@@ -328,6 +328,58 @@ final class FileOperationsTestWindows: XCTestCase {
...
+
+      // Positioned read/write beyond 4 GiB exercises both dwords of the
+      // `OVERLAPPED` offset. Mark the file sparse first so the test does not
+      // allocate several gigabytes of real storage.
+      let handle = try XCTUnwrap(HANDLE(bitPattern: _get_osfhandle(fd.rawValue)))
+      var bytesReturned: DWORD = 0
+      let FSCTL_SET_SPARSE: DWORD = 0x000900C4
+      try XCTSkipUnless(
+        DeviceIoControl(handle, FSCTL_SET_SPARSE, nil, 0, nil, 0,
+                        &bytesReturned, nil),
```
glessard, line 364: Our coding style is to have all the parameters on one line, or to have one line per parameter.
```suggestion
        DeviceIoControl(
          handle, FSCTL_SET_SPARSE, nil, 0, nil, 0, &bytesReturned, nil
        ),
```

glessard (changes requested):
Thanks for this contribution. Could you edit the very wordy comments I reviewed as well as those in the new test to be shorter and more self-contained? There is a reference in the test comment to "the fix", which won't be a useful comment for future readers. "[something about specific values]. Both trapped before the fix." -> "Two values not representable by a 32-bit `off_t`."

Please target the <s>release/1.7.0</s> release/1.8.x branch, as this is clearly a bug fix.

PR #349, opened 2026-07-15
Do not follow reparse points when removing a directory tree on Windows

YayoRazo:
## Summary

`_recursiveRemove` (used by `withTemporaryFilePath` cleanup on Windows) enumerated a directory's children and recursed into any entry with `FILE_ATTRIBUTE_DIRECTORY` set. A directory that is *also* a reparse point — a junction or a directory symlink — carries that attribute, so the routine recursed **through** the link and deleted the contents of its target, which lives outside the tree being removed.

For a temporary directory that happens to contain a junction pointing at, say, the user's documents, this was **silent data loss**: the real files behind the link were deleted, not the link.

## Fix

- Skip reparse points during recursion: remove the link itself with `RemoveDirectoryW`, which deletes the reparse point without touching its target. This matches the POSIX sibling (`FilePathTempPosix.swift`), which only recurses on real directories (`DT_DIR`) and never on symlinks.
- Harden the enumeration in `forEachFile`: `FindNextFileW` returns `false` both at the end of a directory *and* on error, so the loop now verifies it ended with `ERROR_NO_MORE_FILES` rather than treating a transient failure as end-of-directory (which would silently skip the remaining entries and leave the tree partially deleted). `forEachFile` becomes `throws` instead of `rethrows`.

## Testing

- Added `testCleanupDoesNotFollowReparsePoints`: it creates a file in an external "victim" directory, places a **directory junction** to it inside the tree being removed (via `mklink /J`, which needs no special privilege, so the reparse path is exercised in CI), removes the tree, and asserts the victim file survives.
- Verified the test **fails without this change** (the victim file is deleted through the junction) and passes with it.
- Full `swift test` suite green on Windows (`x86_64-unknown-windows-msvc`, Swift 6.3.3); no changes to other platforms.

### DIFF
--- a/Sources/System/FilePath/FilePathTempWindows.swift
+++ b/Sources/System/FilePath/FilePathTempWindows.swift
@@ -35,7 +35,7 @@ internal func _getTemporaryDirectory() throws -> FilePath {
 fileprivate func forEachFile(
   at path: FilePath,
   _ body: (WIN32_FIND_DATAW) throws -> ()
-) rethrows {
+) throws {
   let searchPath = path.appending("\\*")
 
   try searchPath.withPlatformString { szPath in
@@ -59,6 +59,15 @@ fileprivate func forEachFile(
 
       try body(findData)
     } while FindNextFileW(hFind, &findData)
+
+    // FindNextFileW returns false both at the end of the enumeration and on
+    // error; only ERROR_NO_MORE_FILES is the normal terminator. Treating a
+    // transient error as end-of-directory would silently skip the remaining
+    // entries and leave the tree partially deleted.
+    let error = GetLastError()
+    if error != ERROR_NO_MORE_FILES {
+      throw Errno(windowsError: error)
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


Sources/System/FilePath/FilePathTempWindows.swift
```diff
@@ -35,7 +35,7 @@ internal func _getTemporaryDirectory() throws -> FilePath {
 fileprivate func forEachFile(
   at path: FilePath,
   _ body: (WIN32_FIND_DATAW) throws -> ()
-) rethrows {
+) throws {
```
glessard, line 38: This is a spurious change; this should still be `rethrows` until we change `forEachFile()` to throw an `Errno` (or an `Error`-constrained type parameter).

Sources/System/FilePath/FilePathTempWindows.swift
```diff
@@ -73,14 +82,28 @@ internal func _recursiveRemove(
...
-      }
-      let component = FilePath.Component(name)!
-      let subpath = path.appending(component)
+    guard (findData.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) != 0 else {
+      return
+    }
+
+    let name = withUnsafeBytes(of: findData.cFileName) {
+      return SystemString(platformString: $0.assumingMemoryBound(
+                            to: CInterop.PlatformChar.self).baseAddress!)
```
glessard, line 91: Let's replace this with memory rebinding. The older code likely predated the availability of SE-0333.
```suggestion
       $0.withMemoryRebound(to: CInterop.PlatformChar.self) {
         SystemString(platformString: $0.baseAddress!)
       }
```

glessard:
This is indeed a bug, but it only involves an internal utility function. We can take our time.

PR #356, opened 2026-07-20
Update `Stat.generationNumber` to `UInt64`

jrflat:
The underlying `st_gen` is `uint32_t` on Darwin and OpenBSD, but `__uint64_t` on newer FreeBSD versions. Update this API to return `UInt64` instead of `Int` to support all platforms.


glessard (approved):

jrflat:
Verified this passes the FreeBSD checks from https://github.com/apple/swift-system/pull/357

PR #359, opened 2026-07-22
Improve test timing for Android targets

glessard:
It turns out that 10ms is around the clock granularity used by Android's file system, explaining the intermittent failures we've been seeing when expecting the timestamp to increase after a pause. This lengthens the pauses to 50ms (from 10ms), comfortably larger than the clock granularity.


jrflat (approved):
Thanks!

PR #365, opened 2026-07-28
[gardening] improve some import statements

glessard:
libc was implicitly imported in RawIORequests.swift, which will be an error when we can use Swift 6 mode.
Also the `import CSystem` statement in `Syscalls.swift` was inconsistently located compared to other files, where it happens before the platform-specific import.

### DIFF
--- a/Sources/System/IORing/RawIORequest.swift
+++ b/Sources/System/IORing/RawIORequest.swift
@@ -2,7 +2,12 @@
 #if os(Linux)
 
 import CSystem
-    
+#if canImport(Glibc)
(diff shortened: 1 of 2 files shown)


glessard:
@swift-ci please test

jrflat (approved):
LGTM - I have some ideas on how to further rework the imports to be more `internal` and not vend so many `CSystem` types, hope to get a PR up for that soon

glessard:
At this time, the nightly-6.4.x toolchain for Windows is defective and we can't successfully test that configuration.

PR #367, opened 2026-07-29
Prepare for Swift 6 language mode

glessard:
The newer compilers allow us to remove the `@_implementationOnly import` statements, and at the same time get rid of many warnings when building on Linux. This branch contains many changes that compile correctly in Swift 6 language mode, but the swift-ci macOS runners cannot successfully compile the package in that mode. Once they're updated to macOS 15.4 or later, we should be able to enable Swift 6 language mode.

### DIFF
--- a/Sources/System/CMakeLists.txt
+++ b/Sources/System/CMakeLists.txt
@@ -23,8 +23,27 @@ add_library(SystemPackage
   WinSDK+Overlay.swift)
 set_target_properties(SystemPackage PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
+
+# Availability macro definitions.
+# Should stay in sync with the `availability` array in Package.swift,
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -144,5 +144,5 @@ let package = Package(
       dependencies: ["SystemPackage"],
       swiftSettings: [.enableUpcomingFeature("MemberImportVisibility")]),
   ],
-  swiftLanguageModes: [.v5]
```
owenv, line 147: The CMake build probably needs to be updated to match this to prevent incompatibilities from sneaking in and breaking nightly toolchains

glessard: Addressed in [b9be3c1](https://github.com/apple/swift-system/pull/367/commits/b9be3c14a4abf1c46f139bf58a3a7b3effa55401)

glessard:
@swift-ci please test

glessard:
@swift-ci please test

glessard:
At this time, the nightly-6.4.x toolchain for Windows is defective and we can't successfully test that configuration.

PR #369, opened 2026-07-30
Enable Swift 6 language mode

glessard:


glessard:
<s>At this time, the nightly-6.4.x toolchain for Windows is defective and we can't successfully test that configuration.</s>
The August 1st nightly-6.4.x toolchain for Windows finally fixed this.

jrflat:
macOS toolchain build is failing with:
```
MachPort.swift:65:43: error: reference to var 'mach_task_self_' is not concurrency-safe because it involves shared mutable state
```
for all uses of `mach_task_self_`. I'll look into this further.

jrflat:
The toolchain is building with Xcode 16.2 / macOS 15.2 SDK, which is before the `__swift_nonisolated_unsafe` attribute was added to `mach_task_self_`.

jrflat:
We could add a C shim like:

```
static inline mach_port_t _system_mach_task_self(void) { return mach_task_self_; }
```

or trap once and cache the task in Swift:

```
@usableFromInline internal let _machTaskSelf: mach_port_t = task_self_trap()
```

but we could leave it be. I don't _think_ any clients will hit this due to the 6.1 tools-version floor, and `swift-system` won't be pinned to `release/1.8.x` yet in `swift`'s `update-checkout-config.json`, so this shouldn't impact any current toolchain builds. We would need to wait for Swift CI's x86 macOS toolchain jobs to move off of the macOS 15.2 SDK, though, before moving that pin.

jrflat (approved):
Approving this as long as we're OK with waiting for Swift CI bots to move off macOS 15.2 SDK before updating `swift-system`'s toolchain pin in `update-checkout-config.json`.

PR #377, opened 2026-08-08
Recognize non-letter Windows drive designators

### LINKED ISSUE #271: Non-Letter Drive Windows Paths Are Incorrectly Reported As Non-Absolute
This is an edge-case for sure.

Here’s the source code:
```swift
print(FilePath(#"C:\foo"#).isAbsolute)
print(FilePath(#"+:\foo"#).isAbsolute)
print(FilePath(#"€:\foo"#).isAbsolute)
```

The three paths are valid absolute paths on Windows, however the output of this program will be `true`, `false`, `false`.

For more info see <https://www.ryanliptak.com/blog/windows-drive-letters-are-not-limited-to-a-z/>.

aryansk:
Fixes #271

### What changed

- Recognize any single non-separator, non-colon character as a Windows drive designator instead of restricting the syntax to ASCII letters.
- Add simulated-Windows coverage for `+:\\foo`; keep the Unicode `€:` case gated to native Windows where the platform character type supports it.

### Validation

- `swift test --filter FilePathSyntaxTest` — 6 tests passed.

### DIFF
--- a/Sources/System/FilePath/FilePathWindows.swift
+++ b/Sources/System/FilePath/FilePathWindows.swift
@@ -85,10 +85,10 @@ struct _Lexer {
     slice._eat(.backslash) != nil
   }
 
-  // Try to consume a drive letter and subsequent `:`.
+  // Try to consume a drive designator and subsequent `:`.
   mutating func eatDrive() -> SystemChar? {
     let copy = slice
-    if let d = slice._eat(if: { $0.isLetter }), slice._eat(.colon) != nil {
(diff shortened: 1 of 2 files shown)


glessard:
`FilePath` is being moved to the standard library ([SE-0529](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0529-filepath-in-stdlib.md)). Consider contributing there once the change has landed.

aryansk:
Thanks for the heads-up. I’ll revisit the drive-designator behavior in the standard-library FilePath implementation once SE-0529 lands; this draft remains scoped to the current `swift-system` implementation.

PR #379, opened 2026-08-11
fix(MachPort): silence concurrency check for mach_task_self_

### LINKED ISSUE #373: 1.8.0 `mach_task_self_` concurrency build error on macos-14 GHA runners
On macos-14 GHA runners, building swift-system 1.8.0 as a dependency for a SwiftPM 6.3 project reports 10 of the following error in `.build/checkouts/swift-system/Sources/System/MachPort.swift` (I can supply the log, if necessary):

```text
error: reference to var 'mach_task_self_' is not concurrency-safe because it involves shared mutable state
```

I set up the Swift 6.3 toolchain using:

```shell
swiftly init -ny
git reset --hard HEAD # to undo unnecessary changes to .swift-version
swiftly link latest
```

I do not know if other GHA runners also report the same error (or other errors), because the macos-14 failure canceled the builds on other runners.

I do not get any errors building with `swift` from Xcode 26.6 on my macOS 26.6.1 M4 Max Studio.

aryansk:
## Summary

Import Darwin.Mach with `@preconcurrency` so `mach_task_self_` (a shared global from Mach) does not trigger Swift 6.3's concurrency check:

```
error: reference to var 'mach_task_self_' is not concurrency-safe because it involves shared mutable state
```

The existing call sites (`mach_port_guard`, `mach_port_destruct`, etc.) keep their current form; the import attribute suppresses the diagnostic for this known thread-safe global without per-call-site workarounds.

## Issue

Fixes #373 — 1.8.0 `mach_task_self_` concurrency error on macos-14 GHA with Swift 6.3.

## Validation

- `git diff --check` passes

### DIFF
--- a/Sources/System/MachPort.swift
+++ b/Sources/System/MachPort.swift
@@ -9,7 +9,11 @@
 
 #if SYSTEM_PACKAGE_DARWIN
 
+#if canImport(Darwin, _version: 310)
+import Darwin.Mach
+#else
 @preconcurrency import Darwin.Mach
+#endif
 
 @available(System 1.4.0, *)
 public protocol MachPortRight {}
--- a/Sources/System/MachPort.swift
+++ b/Sources/System/MachPort.swift
@@ -9,7 +9,7 @@
 
 #if SYSTEM_PACKAGE_DARWIN
 
-import Darwin.Mach
+@preconcurrency import Darwin.Mach
 
 @available(System 1.4.0, *)
 public protocol MachPortRight {}


glessard:
The concurrency annotations for `mach_task_self_` were introduced in the macOS 15.4 SDK. Is there a way to satisfy the older deployment target without weakening the import for the modern SDKs?

aryansk:
Thanks for the clarification. I checked the issue and the current change. Since the annotations come from the macOS 15.4 SDK, the preconcurrency import would suppress them for modern SDKs too. I do not see a portable SDK-version conditional at this import site, so I am leaving the code unchanged rather than broadening suppression. If Swift System has a supported wrapper or conditional pattern for this SDK boundary, I can apply that.

glessard:
I'll think whether there is another option. Maybe `@preconcurrency` is okay until the lowest addressable deployment target gets past the 2025 OSes, but obviously that's quite some time in the future.

aryansk:
Thanks for considering the SDK and deployment-target tradeoff. I will leave #379 unchanged while you decide whether @preconcurrency is acceptable; I will not broaden the suppression without a supported pattern.

glessard:
We should be able to use a versioned `canImport`, as used here: https://github.com/apple/swift-system/blob/6dc13544bb15f8036f04d3d4f146014967eaedda/Sources/System/FileSystem/Stat.swift#L79

We'll need to find the version of Darwin that came with the original macOS 15.4 SDK.

Sources/System/MachPort.swift
```diff
@@ -9,7 +9,7 @@
 
 #if SYSTEM_PACKAGE_DARWIN
 
-import Darwin.Mach
+@preconcurrency import Darwin.Mach
```
glessard, line 15:
```suggestion
#if canImport(Darwin, _version: 310)
import Darwin.Mach
#else
// for SDKs older than Xcode 16.3
@preconcurrency import Darwin.Mach
#endif
```
At the moment this project doesn't have a test runner configuration to test such an `#else` branch.

aryansk: Implemented the versioned canImport branch in a6582c0: Darwin imports normally for SDK version 310 and newer, while older SDKs retain @preconcurrency. The MachPort test suite passes all 14 tests.

aryansk:
Thanks for the versioned canImport suggestion. I implemented it in a6582c0: Darwin.Mach is imported normally for SDK version 310 and newer, with @preconcurrency retained for older SDKs. swift test --filter MachPort passes all 14 tests.

glessard:
@aryansk Thanks for checking. Could you also target the 1.8.x branch?

aryansk:
Done — I retargeted the PR to `release/1.8.x` as requested. The versioned import fix remains at a6582c0, and the 14-test MachPort suite passes.

glessard:
@aryansk You can mark this ready for review. Out of curiosity, what type of software do you build using an old Xcode/SDK coupled with a much more recent compiler downloaded from swift.org? That combination isn't one we considered when setting up our test matrix.

aryansk:
@glessard Thanks — this came from testing Swift package/library code that needs to retain older macOS deployment compatibility while using a newer swift.org toolchain. The versioned `canImport(Darwin, _version: 310)` branch is now on `release/1.8.x`, and `swift test --filter MachPort` passes all 14 tests. I’m marking the PR ready for review.

